PR #50, opened 2026-03-19
Record: Sliding Window Eval (stride=64), val_bpb=1.1925

mattqlf:
## Summary

- Sliding window evaluation with stride=64 on the baseline 9x512 SP-1024 architecture
- **val_bpb: 1.1925** (post-quant int8+zlib), improving on the Naive Baseline's 1.2244 by **0.032**
- Training is identical to the baseline; the improvement comes entirely from the evaluation strategy
- Each token is scored with 960+ tokens of context instead of 0-1023
- Eval takes 70s on 8xH100 (well within the 10-minute eval budget)
- Total artifact size: 15,874,829 bytes (under 16MB cap)

## Test plan

- [x] Trained and evaluated on 8xH100 SXM (RunPod)
- [x] `final_int8_zlib_roundtrip_exact val_bpb:1.19250007`
- [x] Artifact size verified under 16,000,000 bytes
- [x] `train_gpt.py` compiles and runs within the records folder


0hq (approved):
Looks great!

Issue #17, opened 2026-03-18
Environment Clarifications (pytorch, CUDA, H100)


mikeapedia:
1. The requirements.txt lists torch==2.10.0 but the official runpod image template has pytorch 2.9.1 (CUDA 12.8). Can you please clarify which versions of python, pytorch, and CUDA will be used in the evaluation environment?

2. Will you please update the readme to clarify if you will be using SXM or NVL H100s?

PR #62, opened 2026-03-19
WIP: Add adaptive eval-time context non-record MLX submission

stpcoder:
## Summary

This PR adds a non-record MLX submission under `records/track_non_record_16mb/` for adaptive eval-time context.

The setup is a local Apple Silicon run, not a leaderboard claim. The main change in this snapshot is the final evaluation path: instead of using one fixed policy over the whole validation stream, it does a coarse pass first and then rescoring with a finer stride on the harder windows from that pass.

## Included files

- `README.md`
- `submission.json`
- `train.log`
- `compare_standard.log`
- `train_gpt.py`

## Run details

- hardware: Apple M4 Pro, 48 GB unified memory
- model: SP-1024, `9x512`, `KV4`, tied embeddings
- train data: first FineWeb train shard
- validation: first `32768` validation tokens
- training length: `200` iterations
- adaptive eval settings: `coarse_stride=256`, `fine_stride=64`, `hard_fraction=0.25`

## Result in `train.log`

- pre-quant eval at stop: `val_bpb=2.4070`
- final int8+zlib roundtrip: `val_bpb=2.40284524`
- total submission size: `11297911` bytes

## Same-setup reference

`compare_standard.log` uses the same setup with standard final evaluation:

- standard final roundtrip: `val_bpb=2.41303630`
- adaptive final roundtrip: `val_bpb=2.40284524`

So in this local fixed-step proxy, the adaptive pass improves the final roundtrip score by about `0.01019 bpb`, but it also makes the final eval pass slower. That tradeoff is the reason this is being submitted as a WIP non-record result rather than as a performance claim.

## Validation

- `python -m py_compile records/track_non_record_16mb/2026-03-19_AdaptiveEvalContext_MLX_M4Pro_sp1024_200it/train_gpt.py`
- included local MLX run logs for both adaptive and standard final eval under the same setup

### DIFF
--- a/records/track_non_record_16mb/2026-03-19_AdaptiveEvalContext_MLX_M4Pro_sp1024_200it/README.md
+++ b/records/track_non_record_16mb/2026-03-19_AdaptiveEvalContext_MLX_M4Pro_sp1024_200it/README.md
@@ -1,0 +1,57 @@
+This folder captures a non-record local MLX submission for adaptive eval-time context.
+
+The idea in this snapshot is simple: do one coarse pass over the validation stream, mark the harder windows from that pass, then rescore only those windows with a finer stride. The training setup stays close to the baseline MLX path; the change is in how the final roundtrip evaluation spends extra context.
+
+This is not a leaderboard claim. It is a local Apple Silicon result meant to document the idea, the code snapshot, and a same-setup comparison against standard final evaluation.
+
+Configuration:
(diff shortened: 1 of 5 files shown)


MatoTeziTanka:
### Community Review — WIP: Add adaptive eval-time context non-record MLX submission

```
```

---

PR #86, opened 2026-03-19
Update: 11L MLP3x + WD=0.04 + zstd-22 (val_bpb 1.1502)

aruniyer:
The README content covers everything: 5-seed results, t-stat, methodology


cocohearts:
@aruniyer maintainer repro note: we are doing a pass over merged leaderboard rows and I cannot currently reproduce the claimed `1.1502` record from the files in this PR.

The issue appears to be an internal record mismatch:
- README/submission claim the updated config is 11L / MLP3x / int6 QAT, `model_params=26,501,720`, mean `val_bpb=1.15015359`, around `10,070` steps in 600s.
- The included `train.log` does not match that updated config: it is a 21.8M-param older config and ends at `final_sliding_window_exact val_bpb=1.16660881`.
- Running the included `train_gpt.py` with the README env does produce the 26.5M-param model, but on our exact audit setup it only gets about `4,068-4,090` steps in 600s and the 3-seed mean is `1.18637676`, not `1.15015359`.

Could you reply here with the exact source/logs for the claimed 11L MLP3x run? Ideally we need the per-seed logs that show the claimed `~1.1502` sliding scores and the code/env that produced the ~10k-step throughput. If the current record accidentally contains a stale `train.log` or stale script, please push the corrected record bundle so we can rerun it.

PR #110, opened 2026-03-19
Submission: Top-Heavy FFN Allocation + Packed Int6 Export | pending eval

mr-ashish-panday:
### Approach

This submission focuses on **parameter allocation**, not just export format. The current public frontier is clustered around a uniform `9 x 512` backbone with stronger quantization and evaluation tricks. I kept the proven surrounding recipe, but replaced the uniform `3x` FFN with **OpenELM-style layer-wise scaling** so later blocks get larger FFNs and earlier blocks get smaller ones at the same overall budget.

The exact FFN schedule is:

`768, 960, 1152, 1344, 1536, 1728, 1920, 2112, 2304`

The motivation is that, under a strict parameter budget, later layers are closer to the next-token loss and should benefit more from extra capacity than early layers. This is especially attractive in the Parameter Golf setting because the challenge is strongly **parameter-limited** relative to the 10-minute 8xH100 compute budget. Chinchilla-style reasoning says the available training compute could support a much larger model than the 16 MB artifact allows, so the right move is to spend the byte budget as efficiently as possible inside the model rather than leaving it uniformly distributed.

To support that architecture cleanly, I also implemented an **exact packed-int6 export path** with per-row fp16 scales. This keeps the submission self-contained and avoids relying on an external `zstd` dependency at evaluation time. I keep the tied embedding in fp16 because that tensor is especially quantization-sensitive and serves as both the input embedding and output projection.

Finally, the trainer includes **sliding-window evaluation** and a **CPU `DRY_RUN=1` mode** so the full train/export/eval pipeline can be verified locally without CUDA. Because I do not currently have GPU credits, I am not claiming a score yet and am leaving final eval metrics pending.

### Key Design Decisions

- Use **top-heavy FFN allocation** instead of a uniform `3x` FFN so later layers get more of the parameter budget.
- Keep the proven **9-layer, 512-dim, tied-embedding, GQA** backbone for a controlled comparison against public baselines.
- Export large matrices as **packed int6 + per-row fp16 scale + zlib** to make the artifact fully self-contained.
- Keep `tok_emb.weight` in **fp16** because it is the most quantization-sensitive tensor in public submissions.
- Retain **sliding-window eval** because it is allowed by the rules and is consistently helpful in public runs.
- Use **seed 42** by default and add a **10-step CPU dry-run** path for reproducibility without GPUs.

### What I Expect to Underperform / Risks

- The main risk is that layer-wise FFN scaling is theoretically motivated but not yet validated on this exact challenge objective.
- Packed int6 + zlib is self-contained, but the trained checkpoint still needs to prove it compresses comfortably under the 16 MB cap.
- If this challenge is dominated more by export/eval tricks than by parameter placement, the gain over a strong uniform-FFN baseline may be modest.

### Results

| Metric | Value |
|---|---|
| Baseline eval loss | `1.22436570` val_bpb (repo baseline) |
| Final eval loss | Pending — training to be run once compute credits are granted |
| Parameter count | `21,778,504` |
| Artifact size | Measured init-export: `4,273,390` bytes total; dense-random stress probe: `16,549,133` bytes total |
| Training time (estimated on 8×H100) | Expected to fit the 600s budget; public 21.8M-parameter 9x512 runs report roughly `45–57 ms/step`, implying about `10k–13k` steps in 10 minutes |

### Reproducibility

Seed: `42`.

Dry-run verified locally with:

`DRY_RUN=1 RUN_ID=topheavy_dryrun python train_gpt.py`

Full 8xH100 train/eval pending compute access.

### References

- Hoffmann et al., *Training Compute-Optimal Large Language Models*, 2022
- Mehta et al., *OpenELM*, 2024

### DIFF
--- a/notes.md
+++ b/notes.md
@@ -1,0 +1,164 @@
+# Parameter Golf Notes
+
+## Repo Rules
+
+### Exact 16 MB Artifact Definition
+
+- The hard cap is **decimal** `16,000,000` bytes, not `16 MiB`.
+- The counted artifact is:
+  - `len(train_gpt.py in UTF-8 bytes)`
+  - plus the **compressed model bytes**
+- The README is explicit that counted code should live in `train_gpt.py`.
+- No network calls, dataset access, or external downloads are allowed during evaluation.
+- Evaluation itself also has a separate runtime cap: under 10 minutes on `8xH100`.
+
+Source:
+- `README.md`, FAQ section and submission process.
+
+## Exact Eval Metric
+
+From `train_gpt.py`:
+
+1. Validation loss is average token cross-entropy in **natural log units**:
+   - `val_loss = sum(token_ce) / num_val_tokens`
+2. Convert to bits per token:
+   - `bits_per_token = val_loss / ln(2)`
+3. Compute tokenizer-aware bytes represented by each target token:
+   - base UTF-8 byte length of the SentencePiece token
+   - plus one extra byte for a leading space when the piece starts with `▁` and the previous token is not a boundary token
+4. Convert to tokens per byte:
+   - `tokens_per_byte = num_val_tokens / num_val_bytes`
+5. Final metric:
+   - `val_bpb = bits_per_token * tokens_per_byte`
+
+Important implementation detail:
+- Baseline validation uses **non-overlapping** chunks of length `TRAIN_SEQ_LEN` (or `EVAL_SEQ_LEN` in some PRs).
+- Many strong PRs improve score further with **sliding-window eval**, but the underlying `val_bpb` formula above stays the same.
+
+Relevant code:
+- `train_gpt.py`: `build_sentencepiece_luts`, `load_validation_tokens`, `eval_val`
+
+## Baseline Architecture
+
+Baseline config from README, baseline record, and root `train_gpt.py`:
+
+- Decoder-only GPT
+- `VOCAB_SIZE=1024`
+- `NUM_LAYERS=9`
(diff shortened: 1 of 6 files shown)


MatoTeziTanka:
### Community Review — Submission: Top-Heavy FFN Allocation + Packed Int6 Export | pending eval

**BPB:** (not parsed — see PR title) | **Compliance:** LOOKS CLEAN — pure-neural submission, no TTT/SLOT/n-gram-cache

**What I found in the code** (head SHA `411d753b1c4e`, file `records/track_10min_16mb/2026-03-19_TopHeavyFFN_PackedInt6_PendingEval/train_gpt.py`):

**CPU smoke test (CT2038 proteus-engine, 2026-04-11):** import OK in 0.02s, dim=512, layers=9, vocab=1024, code=60837 B, SMOKE_TEST_PASS

**Verdict:** LOOKS CLEAN.

---
*Reviewed by [@MatoTeziTanka](https://github.com/MatoTeziTanka) — [The Agora](https://matotezitanka.github.io/parameter-golf). CPU smoke test (CT2038 proteus-engine, 2026-04-11): import OK in 0.02s, dim=512, layers=9, vocab=1024, code=60837 B, SMOKE_TEST_PASS. Classification via deterministic AST-based `classify_prs.py` (pattern bank derived from ~65 manually-reviewed PRs earlier in the 2026-04-11 sweep). This review was auto-drafted from a template and spot-checked before posting — if the template misread your code, please call it out so I can iterate the classifier.*

PR #119, opened 2026-03-19
Add Psionic non-record runtime replay submission

AtlantisPleb:
This PR adds one non-record submission folder generated from Psionic's canonical export path:

- `records/track_non_record_16mb/2026-03-18_psionic_local_reference_runtime_replay_v2`

Claim posture:
- non-record only
- no claim of record-track readiness
- no claim of true `8xH100` success

Local verification before opening this PR:
- `scripts/check-parameter-golf-record-folder-compatibility.sh --parameter-golf-root ~/code/parameter-golf --submission-dir ...` -> `compatible`
- `scripts/check-parameter-golf-record-folder-replay.sh --submission-dir ...` -> `verified`
- folder-local `python3 train_gpt.py` dry-run exit code `0`

The shipped folder includes the submission surface plus the replay-verification, run-evidence, and promotion-receipt artifacts produced by Psionic's current non-record path.

Maintainer-facing note:
- this submission uses a shipped Rust runtime payload launched by the root-local `train_gpt.py` wrapper
- the current Psionic repo still treats counted-runtime and build-dependency interpretation for record-track posture as an open question, so this PR is intentionally non-record and bounded


AtlantisPleb:
Maintainer-facing accounting question from the Psionic side for later record-track posture:

This non-record folder ships a root-local `train_gpt.py` launcher plus a prebuilt Rust runtime payload inside the submission folder, and it requires no build step or network fetch during execution. For future record-track interpretation, is the intended accounting posture simply "count the shipped bytes exactly as they exist in the folder" including that runtime payload, or is there any different expectation for self-contained prebuilt helper/runtime binaries versus source-only submissions?

We are already counting the shipped runtime bytes explicitly on the Psionic side, but we want the upstream interpretation preserved on-record rather than inferred.

PR #144, opened 2026-03-20
Add MPK 8x384 10-minute submission record

DJLougen:
Add MPK 8x384 10-minute submission record

This PR adds a new 10-minute / 16 MB submission record:
- records/track_10min_16mb/2026-03-20_MPK_8x384

Highlights:
- model_family: mpk
- 8 layers, width 384
- 8 heads / 4 KV heads
- MPK strides k=2, m=4
- 80 FineWeb SP-1024 train shards
- wallclock stop at 599.527s
- final int8+zlib roundtrip val_bpb: 1.01558307
- total submission size int8+zlib: 14578273 bytes

Included files:
- README.md
- submission.json
- train.log
- train_gpt.py


andrewgcodes:
Really interesting approach! I was wondering about [line 205](https://github.com/openai/parameter-golf/pull/144/changes#diff-49393f3b459b944faae34a2b572e561f81f7e4c156a5ebdd6befcb97f22a48b7R205) of train_gpt.py. Could this have an effect on val_bpb?

DJLougen:
Yep, that definitely could affect val_bpb — thanks for catching it. I confirmed the line in the committed source and I’m retesting with the corrected SentencePiece leading-space marker handling now. I’ll post the corrected logs/results as soon as the rerun finishes.

DJLougen:
﻿I confirmed the bug and pushed a correction.

The issue was in `build_sentencepiece_luts`: the SentencePiece leading-space marker check was committed as `piece.startswith("?")` instead of `piece.startswith("?")`, which invalidated the previously reported BPB values.

I've now updated the record folder to:
- fix the marker check in the submitted `train_gpt.py`
- add a corrected rerun log: `mpk_seed_1341_bpbfix.txt`
- update `README.md` and `submission.json` to reflect the corrected numbers

Sadly the corrected rerun result:
- `final_int8_zlib_roundtrip_exact val_loss:2.28232567 val_bpb:1.35172182`

So the earlier `~1.01` BPB claim should be disregarded; it was due to the accounting bug above. Thanks again for catching it.

PR #152, opened 2026-03-20
Add TTT (Test-Time Training) submission: 1.1767 BPB

timowhite88:
Full-model SGD adaptation during eval phase improves BPB by 3.0% over static inference with zero architecture changes.


leloykun:
Hi @timowhite88 ! Are you certain you're not leaking future tokens during your TTT adaptation? From the looks of it, `epochs=1` already leaks information as you do it _before_ doing any evals, not during evals as you go. `epochs=2` seems to make it worse.

timowhite88:
Hey @leloykun — no leakage. TTT adaptation uses causal masking

leloykun:
Hmmm... I'm hoping I'm not sounding too critical here. I was actually one of the speedrunners in the original `modded-nanogpt` repo, and we had a lot of convos like this back then too.

That said, no, this is still leakage. Even when we're evaluating those compression algorithms, we still typically don't allow them to use statistics from the 'hidden' validation set. At most, we only allow them to update their 'cache' online only on information they've already 'seen' so far. And besides, if the goal is to just compress both the training and validation sets, why don't we just use gzip? It's cheaper and lossless.

I also want you to look at this from a practical perspective during inference: even if the model is getting fed with external information (from, say, camera feeds of a self-driving car), the model still cannot use information past time `t`! It can only adapt to the distribution of the things it has seen so far.

So, the non-leaky version of TTT goes something like:
1. Adapt to information at time `t-1` (and backwards);
2. Do inference at time `t`;
3. Score predictions at time `t`;
4. Repeat.

Wdyt @0hq ?

PR #189, opened 2026-03-20
Harden TTT doc splitting for missing/trailing BOS edge cases

shuofengzhang:
## What changed
- Hardened `_find_docs` in `train_gpt.py` for tokenizer/export edge cases:
  - If no BOS tokens are present, it now falls back to a single document covering the full validation token stream (when length >= 2).
  - It now skips degenerate BOS-only fragments (`doc_len < 2`) instead of asserting.
  - It now raises explicit `ValueError`s when there are no valid validation docs.
- Added focused regression tests in `tests/test_find_docs.py` for:
  - no-BOS fallback behavior,
  - trailing BOS-only fragment handling,
  - invalid too-short sequence error path.

## Why
- The script is intended to be tokenizer-agnostic, but `_find_docs` previously assumed BOS boundaries always exist and all BOS-delimited docs have at least two tokens.
- In edge cases (custom tokenizers, export quirks, trailing BOS-only shard tails), this could trigger assertion failures or downstream divide-by-zero-style failures in TTT validation.

## Insight / Why this matters
- **Root cause:** document segmentation relied on strict BOS invariants that are true for the common path but not guaranteed for all tokenizer/export variants.
- **Why this is easy to miss:** normal challenge datasets typically include BOS and non-degenerate docs, so failures appear only in less common validation inputs.
- **Tradeoff:** the fallback/skip logic favors robust evaluation continuity over strict hard-fail behavior for every malformed boundary.
- **Impact:** evaluation becomes more resilient and fails with clearer messages when data is truly invalid, reducing wasted long training runs that otherwise crash at validation time.

## Practical gain / Why this matters
- Users running custom tokenizers or modified preprocessing pipelines get a robust TTT validation path instead of brittle assertion crashes.
- Maintainers get deterministic regression coverage for boundary conditions that were previously untested.

## Testing
- ✅ `pytest -q tests/test_find_docs.py`
- ✅ `pytest -q`

## Risk analysis
- Scope is intentionally narrow: only document-boundary parsing and new targeted tests.
- No changes to model architecture, optimizer math, or training loop scheduling.
- Behavior for standard BOS-formatted datasets remains unchanged.


MatoTeziTanka:
# PR #189 Review

**Title:** Harden TTT doc splitting for missing/trailing BOS edge cases
**State:** open
**Date Reviewed:** 2026-04-11

## Code Analysis

### train_gpt.py Checks
- **target-in-key pattern:** not found
- **TTT (Temporal Token Tagging):** not found
- **SLOT (Slot MoE):** not found
- **Custom Tokenizer:** not found

## Verdict

**Classification:** PURE_NEURAL_CLEAN

**Recommendation:** 

**Recommendation to @cocohearts @valerio-oai @0hq @yuzhougu-oai @notapplica:** ✓ MERGE

---
Draft by @MatoTeziTanka for parameter-golf review sweep (2026-04-11)

---
*Reviewed by [@MatoTeziTanka](https://github.com/MatoTeziTanka) — [The Agora](https://matotezitanka.github.io/parameter-golf). Classification via sibling-session agent (Haiku-backed). This review was drafted from a template and spot-checked before posting — if the template misread your code, please call it out so I can iterate the classifier.*

PR #323, opened 2026-03-21
Minimal recurrent motif (sb1 rs2 g0.18) – non-record submission

megnat05-tmm:
This is a non-record submission.

## Summary

This submission introduces a minimal recurrent motif architecture that achieves improved compression under the 16MB constraint by emphasizing structural reuse over explicit depth.

The model uses a single shared block (`shared_block_size=1`) with limited recurrence (`recurrence_steps=2`) and soft gating (`recurrence_gate_init=0.18`). This design was motivated by the idea that large effective structures can be generated through a compact operator rather than stored explicitly.

## Results

Final (roundtrip):
- val_loss: 4.7062
- val_bpb: 2.7873

Artifact:
- compressed size: ~1.92 MB
- raw size: ~8.47 MB

This configuration outperformed larger motif variants in both compression and efficiency.

## Approach

The architecture explores recurrence as a structural closure mechanism. A compact shared operator is reused across steps to generate extended representations. This reduces parameter requirements while preserving model capacity.

## Notes

- Validation timing on local hardware reflects evaluation chunking and logging cadence, but does not affect correctness of metrics.
- The submission is fully self-contained and reproducible.


MatoTeziTanka:
> **[RETRACTED 2026-04-11]** — This IMPORT_FAIL was a false positive. Root cause: runner fetched a path marked deleted in the PR diff. Your code is not broken. See correction below: https://github.com/openai/parameter-golf/pull/323#issuecomment-4230238998

---

### Community Review — Minimal recurrent motif (sb1 rs2 g0.18) – non-record submission

```
SyntaxError: (unicode error) 'utf-8' codec can't decode byte 0x9e in position 0: invalid start byte (line 1)
```

---
*Reviewed by [@MatoTeziTanka](https://github.com/MatoTeziTanka) — [The Agora](https://matotezitanka.github.io/parameter-golf). CPU smoke test (CT2038 proteus-engine, 2026-04-11): **IMPORT_FAIL** — SyntaxError: (unicode error) 'utf-8' codec can't decode byte 0x9e in position 0: invalid start byte (line 1). Classification via `classify_prs.py` AST-based classifier; full compliance audit deferred until the import issue is resolved. Auto-drafted from a template and spot-checked before posting.*

MatoTeziTanka:
### Retraction — this IMPORT_FAIL was a deleted-file artifact in my smoke runner

Sorry @megnat05-tmm, this one's on me. I re-audited the `SyntaxError (unicode error) 'utf-8' codec can't decode byte 0x9e in position 0` I reported above and it was a false positive — the fault is in my smoke runner, not in your code.

**What happened:**

Your PR deletes 16 old `records/*/train_gpt.py` path(s) while editing a different file, and my bulk smoke runner iterated the diff's file list and fetched one of the paths that's already marked for deletion. The raw GitHub content endpoint returned either a binary stub or a non-UTF8 response, and my runner tried to import it as Python source, producing the `byte 0x9e at position 0` error. That error was about the deleted/non-existent file, not the `train_gpt.py` you're actually submitting.

**Verified at head `bff5e2d`:**

The real `train_gpt.py` you're editing parses cleanly under Python 3.10:

```
py_compile.compile('train_gpt.py') → PARSES OK
71409 bytes
```

**Your PR is not broken by this error.** I'm retracting the IMPORT_FAIL classification. I'll re-queue the full compliance audit and post findings separately.

Again — sorry for the noise. I'm adding a "don't fetch paths marked deleted in the PR diff" guard to the runner so this doesn't hit other PRs that delete/rename records folders.

PR #523, opened 2026-03-23
Add RIOM v1 shared-depth recurrence non-record submission

hesong0222-dev:
## Summary

This draft PR adds a non-record RIOM-style submission under `records/track_non_record_16mb/2026-03-23_RIOM_v1_recur`.

The core idea is shared-depth recurrence: replace 9 distinct transformer blocks with 3 shared blocks repeated 3 times, with lightweight learned recurrence gates. The tokenizer and dataset remain unchanged from the official SP1024 FineWeb setup, and the change is isolated to the record-local `train_gpt.py`.

## What actually ran

This package is based on a real Apple Silicon MLX development run.

- training steps: `50`
- validation prefix: `1,048,576` official validation tokens
- pre-quant: `val_loss=5.4143`, `val_bpb=3.2439`
- post-quant: `val_loss=5.42207763`, `val_bpb=3.24862008`
- code bytes: `51,404`
- compressed model bytes: `2,273,437`
- total counted bytes: `2,324,841`

## Why this is interesting even if non-record

On the same local development prefix, this recurrent package improved `val_bpb` over the local baseline package while sharply reducing counted artifact size. That is the main RIOM hypothesis in compact form: more effective depth per byte through parameter sharing.

## Limitations

- This is not a record claim.
- This is a local MLX development run, not an 8xH100 submission.
- Validation was capped with `VAL_MAX_TOKENS`; it should be rerun with `VAL_MAX_TOKENS=0` before any serious upstream submission.
- The next concrete step is a CUDA/PyTorch port into the official `train_gpt.py` path.


MatoTeziTanka:
### Community Review — Add RIOM v1 shared-depth recurrence non-record submission

```
```

---

hesong0222-dev:
Thanks for the smoke test.

I pushed a fix for the import blocker on the RIOM v1 non-record script. The update guards the MLX and `sentencepiece` imports so the module can be imported on a machine without those packages installed, while still failing clearly at runtime if someone actually tries to run the MLX path without the required dependencies.

What I verified locally after the change:

- `python3 -m py_compile records/track_non_record_16mb/2026-03-23_RIOM_v1_recur/train_gpt.py`
- direct module import in an environment without `mlx` and without `sentencepiece`
- calling `main()` in that same environment now raises a clear `ModuleNotFoundError` explaining which runtime dependencies are missing, instead of failing during import

If you re-run the compliance pipeline, the import step should now pass cleanly.

hesong0222-dev:
Follow-up: I re-ran the checks under Python 3.10 specifically.

- `python3.10 -m py_compile records/track_non_record_16mb/2026-03-23_RIOM_v1_recur/train_gpt.py` now passes
- direct module import also passes on a machine without `mlx` and without `sentencepiece`
- in that same environment, calling `main()` now fails cleanly with a runtime dependency error instead of failing during import

PR #542, opened 2026-03-23
Non-Record: DG Attention, Differential-Gated Attention with Depth-Scheduled Novelty Encoding: (val_bpb=1.1898)

ddavidgao:
## Summary
 
Novel attention mechanism where deep layers transmit *what's new* about each token instead of raw content, with a depth schedule discovered from empirical β trajectory analysis. Named "DG" for **D**esignator/**G**radient.
 
---
 
## The Idea
 
Standard QKV attention asks three questions per token:
- **Q**: "What am I looking for?"
- **K**: "What do I offer?"
- **V**: "Here's everything about me."
 
DG Attention asks two:
- **D** (Designator): "Who matches me?" — asymmetric D_q/D_k projections for matching, using Flash Attention via `F.scaled_dot_product_attention`.
- **G** (Gradient): "Here's what's *new* about me compared to what you already know." — projects the difference between the current token and a causal running baseline.
 
The core insight comes from video compression: codecs don't send full frames, they send frame *differences*. Similarly, DG's deep layers don't transmit raw token content — they transmit the delta relative to prior context. Early layers still use raw content because they're building initial representations. This depth-dependent behavior was discovered empirically and hardcoded into the architecture.
 
**Distinct from Microsoft's Differential Transformer** (ICLR 2025), which differences *attention weights* across two softmax maps. DG differences the *value payload* against a causal baseline. Different mechanism, different thesis.
 
---
 
## Results
 
| Run | BPB | Steps | ms/step | Artifact |
|---|---|---|---|---|
| Standard attention (baseline) | 1.1558 | 4,451 | 134ms | 18.7 MB |
| DG v2.1 — no Flash Attention | 1.28 | 1,387 | 432ms | 15.9 MB |
| **DG v2.1 — with SDPA** | **1.1898** | **2,979** | **201ms** | **15.9 MB** |
 
**5080 laptop convergence test** (2,000 steps, identical conditions):
 
| Step | Standard | DG v2.1 | Delta |
|---|---|---|---|
| 200 | 2.334 | 2.331 | -0.003 |
| 600 | 2.186 | 2.190 | +0.004 |
| 1000 | 1.980 | 2.053 | +0.072 |
| 1400 | 1.899 | 1.934 | +0.035 |
| 1800 | 1.832 | 1.816 | **-0.017** |
| 2000 | 1.821 | 1.795 | **-0.026** |
| **Final** | **1.826** | **1.800** | **-0.026** |
 
DG fell behind during mid-training (steps 400–1200), then overtook standard attention and the gap widened through the end of training. This is not early inductive bias — it's a durable advantage that emerged after the mechanism learned to use the differential pathway.
 
---
 
## The β Finding
 
DG uses a hybrid payload: `payload = β * raw_content + (1 - β) * differential_signal`, where β is learned per layer. Lower β = more differential.
 
**On the 5080 (small batches, 2,000 steps)**, β collapsed toward 0 in deep layers:
 
| Layer | β @ step 500 | β @ step 2000 |
|---|---|---|
| L0 (first) | 0.387 | 0.471 |
| L5 (middle) | 0.578 | 0.591 |
| L10 (last) | 0.038 | **0.009** |
 
The last layer was 99.1% differential. The model deeply committed to novelty encoding in deep layers while keeping raw content in early layers.
 
**On H100 (large batches, 3,000 steps)**, β drifted back toward raw content:
 
| Layer | β @ step 500 | β @ step 2500 |
|---|---|---|
| L0 (first) | 0.642 | 0.758 |
| L5 (middle) | 0.359 | 0.747 |
| L10 (last) | 0.199 | 0.401 |
 
The optimizer, given larger batches and more stable gradients, found it easier to route through raw content — the same gate collapse phenomenon seen in Mixture-of-Experts models.
 
**This motivated the depth schedule in v3**: instead of letting β collapse, hardcode the depth-dependent pattern the data revealed. Early layers get raw content, deep layers get differential, no learned gate to collapse.
 
---
 
## Version History
 
| Version | Changes | Result |
|---|---|---|
| **v1** | Symmetric D (D×D^T), running mean baseline, pure differential | BPB 2.327 — 0.037 worse than baseline |
| **v2.1** | Asymmetric D_q/D_k, parallel cumsum baseline, hybrid payload with learned β | BPB 1.800 — **beats baseline** (1.826) at convergence |
| **v2.2** | Blended baseline (γ × prev_token + (1-γ) × global_mean) | BPB 1.800 — identical to v2.1, proved global mean is sufficient |
| **v3** | Hardcoded depth schedule replacing learned β, Flash Attention throughout | Eliminates gate collapse, reduces overhead by ~20ms/step |
 
Key iteration: every expert criticism from v1 (symmetric matching loses directionality, running mean too blunt, pure differential too rigid) was addressed in v2.1. The convergence test validated the fixes.
 
---
 
## Limitations
 
- **201ms/step vs 80ms for standard.** The cumsum baseline computation and dual payload projections add ~120ms of overhead that isn't covered by Flash Attention. This means DG gets ~3,000 steps in 10 minutes vs ~7,000 for standard — roughly half the training.
- **β trends toward raw content at H100 scale.** The differential pathway loses influence with larger batches. The depth schedule (v3) addresses this, but hasn't been validated on H100 yet.
- **Not tested with TTT, XSA, SwiGLU.** Current SOTA submissions stack these optimizations. DG was tested in isolation against vanilla attention. The BPB gap would likely close or reverse when stacked with the full optimization suite.
- **Artifact size exceeds 16MB limit.** Total submission size is 16,638,468 bytes — 638KB over the 16,000,000-byte cap. This is due to full-width designator projections added for Flash Attention compatibility. Fixable with more aggressive int5 quantization or narrower designators with a custom Triton kernel.
- **26M parameter model only.** Scaling behavior beyond this size is unknown.
 
---
 
## What's Next
 
- **Stack TTT + SwiGLU + XSA on top of DG** — the techniques driving current SOTA should compose with DG since they operate on different parts of the architecture.
- **Depth-scheduled v3 on H100** — validate that hardcoded β eliminates collapse without hurting BPB.
- **Custom Triton kernel for cumsum baseline** — fuse the baseline computation to close the ms/step gap.
- **Investigate batch size → β trajectory relationship** — why do large batches cause gate collapse? This has implications beyond DG for any gated architecture.
- **ArXiv preprint forthcoming** — "DG Attention: Novelty-Conditioned Value Streams for Parameter-Efficient Language Modeling."
 
---
 
## Prior Art & Distinction
 
- **Reformer** (Kitaev et al., ICLR 2020): Explored shared Q/K attention. DG's asymmetric D_q/D_k is related but distinct.
- **Differential Transformer** (Microsoft, ICLR 2025): Differences *attention score maps* to cancel noise. DG differences the *value payload* against a causal baseline. Different mechanism entirely.
- **Shared-Weight Self-Attention** (2024): Showed shared Q/K/V maintains performance with 66% fewer attention params. DG goes further by replacing V with a novelty-conditioned signal.


MatoTeziTanka:
### Community Review — Non-Record: DG Attention, Differential-Gated Attention with Depth-Scheduled Novelty Encoding: (val_bpb=1.1898)

**BPB:** 1.1898 | **Compliance:** LOOKS CLEAN — pure-neural submission, no TTT/SLOT/n-gram-cache

**What I found in the code** (head SHA `c7311ad286bc`, file `records/track_non_record_16mb/2026-03-23_DGAttention_DavidGao/train_gpt.py`):

**CPU smoke test (CT2038 proteus-engine, 2026-04-11):** import OK in 0.08s, dim=512, layers=11, vocab=1024, code=80773 B, SMOKE_TEST_PASS

**Verdict:** LOOKS CLEAN.

---
*Reviewed by [@MatoTeziTanka](https://github.com/MatoTeziTanka) — [The Agora](https://matotezitanka.github.io/parameter-golf). CPU smoke test (CT2038 proteus-engine, 2026-04-11): import OK in 0.08s, dim=512, layers=11, vocab=1024, code=80773 B, SMOKE_TEST_PASS. Classification via deterministic AST-based `classify_prs.py` (pattern bank derived from ~65 manually-reviewed PRs earlier in the 2026-04-11 sweep). This review was auto-drafted from a template and spot-checked before posting — if the template misread your code, please call it out so I can iterate the classifier.*

valerio-oai (approved):

PR #553, opened 2026-03-23
10L + GatedXSA + EMA + TrigramHash + QuadgramHash + SwiGLU (val_bpb=TBD)

Vighaneshs:
PS: I don't have funds for training on 8xh100.

Building on the 10L SOTA (2026-03-20_10L_Int5MLP_MuonWD04_SWA50, 1.14276 BPB). Four additions:

1. Gated XSA — XSA (arXiv:2603.09078) with a learned per-layer sigmoid gate instead of hard projection removal. Starts near full XSA (sigmoid(4.0) ≈ 0.98) and adapts during training. The intuition is that early and late layers may want different amounts of self-value suppression, so letting it learn seems better than fixing it at 1.0.

2. EMA — Exponential moving average (decay=0.997) replacing SWA. Shadow copy on GPU, updated every step. Feels cleaner than SWA since you don't have to pick a start fraction — it naturally tracks the model through the whole warmdown.

3. TrigramHash + QuadgramHash — Hash (t-2, t-1, t) and (t-3, t-2, t-1, t) into learned embedding tables (1024 buckets each, dim=64/32), added at the input. Cheap in terms of compressed bytes (~55KB and ~25KB respectively after int8+zstd) but gives the model explicit n-gram context at zero attention overhead. Uses int32 arithmetic for the hash to avoid overflow.

4. SwiGLU — Swapped relu² for SwiGLU with mlp_mult=2.0 (gate+up+down vs the old fc+proj). Param count stays the same as relu² at mlp_mult=3.0.

BigramHash is back to 10240 buckets (had to shrink it to 2048 in a previous 11L attempt to fit the budget — staying at 10L means we don't have to make that tradeoff).

All weights quantized int5 (MLP) / int6 (attn, bigram, trigram, quadgram proj) / int8 (embedding tables) + zstd-22. Fits within 16MB.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,3 +9,4 @@ data/docs_selected.jsonl
 .mypy_cache/
 .venv
 logs/
+temp/
--- a/records/track_10min_16mb/2026-03-21_11L_XSA4_EMA_TrigramHash/README.md
+++ b/records/track_10min_16mb/2026-03-21_11L_XSA4_EMA_TrigramHash/README.md
@@ -1,0 +1,112 @@
+# 11L + XSA(last 4) + EMA + TrigramHash + Int5MLP + BigramHash + SWA
+
+**val_bpb: TBD** (mean of 3 seeds, sliding window stride=64, post int5/int6+zstd quantization roundtrip)
+
+Three additions over the `2026-03-20_10L_Int5MLP_MuonWD04_SWA50` SOTA (val_bpb=1.14276):
+
+## Run Command
+
+```bash
+# Setup (once)
+bash prepare.sh
+
+# Train + evaluate (default seed=42)
+bash eval/eval.sh
+
+# With specific seed
+SEED=42 bash eval/eval.sh
+SEED=1337 bash eval/eval.sh
+SEED=7 bash eval/eval.sh
+```
+
+All parameters are set as defaults in `train_gpt.py`. No env vars needed.
+
+## 3-Seed Results
+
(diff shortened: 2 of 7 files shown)


MatoTeziTanka:
### Community Review — 10L + GatedXSA + EMA + TrigramHash + QuadgramHash + SwiGLU (val_bpb=TBD)

**BPB:** 1.1428 | **Compliance:** LOOKS CLEAN — pure-neural submission, no TTT/SLOT/n-gram-cache

**What I found in the code** (head SHA `27f8be3ba8a0`, file `records/track_10min_16mb/2026-03-21_11L_XSA4_EMA_TrigramHash/train_gpt.py`):

**CPU smoke test (CT2038 proteus-engine, 2026-04-11):** import OK in 0.03s, dim=512, layers=11, vocab=1024, code=61542 B, SMOKE_TEST_PASS

**Verdict:** LOOKS CLEAN.

---
*Reviewed by [@MatoTeziTanka](https://github.com/MatoTeziTanka) — [The Agora](https://matotezitanka.github.io/parameter-golf). CPU smoke test (CT2038 proteus-engine, 2026-04-11): import OK in 0.03s, dim=512, layers=11, vocab=1024, code=61542 B, SMOKE_TEST_PASS. Classification via deterministic AST-based `classify_prs.py` (pattern bank derived from ~65 manually-reviewed PRs earlier in the 2026-04-11 sweep). This review was auto-drafted from a template and spot-checked before posting — if the template misread your code, please call it out so I can iterate the classifier.*

PR #664, opened 2026-03-25
Non-record: hybrid spiking Transformer (SNN)with a multi-step spiking MLP

tsbiosky:
# Hybrid Spiking Neural Networks (SNNs) MLP

**val_bpb: 1.2982** | **15.78 MB** | 8×H100 SXM

A contest-friendly hybrid SNN submission built from the `train_gpt.py` baseline: keep dense GQA attention and the original training/eval/compression pipeline, but replace the standard  feed-forward block with a small multi-step leaky integrate-and-fire (LIF-style) spiking MLP.

Reference :https://arxiv.org/pdf/2203.14679

## Why this is interesting

This is **not** a fully spiking language model. It is a **hybrid Transformer + SNN-MLP** design:

- embeddings, attention, residual path, and logits remain standard dense LM components
- only the feed-forward block is replaced by a spiking mechanism
- the original Parameter Golf training and export path stays intact

That makes the experiment meaningful for the contest setting because it isolates one question:

> Can spike neural network achieves good performance in a tiny language model under a strict size budget?

### DIFF
--- a/records/track_non_record_16mb/2026-03-24_Hybrid_SNN_MLP/README.md
+++ b/records/track_non_record_16mb/2026-03-24_Hybrid_SNN_MLP/README.md
@@ -1,0 +1,157 @@
+# Hybrid Spiking Neural Networks (SNNs) MLP
+
+**val_bpb: 1.2982** | **15.78 MB** | 8×H100 SXM
+
+A contest-friendly hybrid SNN submission built from the `train_gpt.py` baseline: keep dense GQA attention and the original training/eval/compression pipeline, but replace the standard  feed-forward block with a small multi-step leaky integrate-and-fire (LIF-style) spiking MLP.
+
+Reference :https://arxiv.org/pdf/2203.14679
+
+## Results (8×H100 80GB SXM)
+
+| Run | Val loss | **Val bpb** | Serialized model | int8+zlib model | Total submission |
+|-----|----------|-------------|------------------|-----------------|------------------|
+| SNN baseline | 2.1919 | **1.2982** | 67,233,157 | 15,723,303 | **15,776,086** |
+
+### Exact export log
+
+- Serialized model: **67,233,157 bytes**
+- Code size: **52,783 bytes**
+- Total submission size: **67,285,940 bytes**
+- Serialized model int8+zlib: **15,723,303 bytes**
+  - payload: 17,179,020 bytes
+  - raw_torch: 17,231,715 bytes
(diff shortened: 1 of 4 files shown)


MatoTeziTanka:
### Community Review — Non-record: hybrid spiking Transformer (SNN)with a multi-step spiking MLP

## PR #664 — Non-record: Hybrid SNN-Transformer (tsbiosky / Junxuan Huang)
**Head SHA:** d8a3cfb5bcfb9c276ecb40b17cc8e2e740ccfcff  
**File audited:** `records/track_non_record_16mb/2026-03-24_Hybrid_SNN_MLP/train_gpt.py` (1238 lines)

---

### Check 1 — N-gram family bug (CLOSE trigger)
**Result: NOT PRESENT**

No n-gram, bigram, or hash-lookup mechanism exists anywhere in this file. There are no BigramHash, NgramHash, or similar structures. No token is used as a key in a hash lookup anywhere in the forward pass. The only LUT structures are `base_bytes_lut`, `has_leading_space_lut`, and `is_boundary_token_lut`, which are SentencePiece tokenizer byte-count helpers for BPB computation — these are not n-gram predictors and do not touch model logits. N-gram CLOSE criterion does not apply.

### Check 2 — Pre-Quant TTT (CLOSE trigger)
**Result: NOT PRESENT**

No test-time training of any kind is present. The val path (`eval_val`) is a standard read-only inference loop — no optimizer steps, no backward calls, no AdamW or any optimizer invoked outside the main training loop. The only use of `val_tokens` is in `eval_val()` (inference-only) and the post-quantization round-trip validation (also inference-only). There is zero gradient flow through val_tokens at any point. Pre-Quant TTT CLOSE criterion does not apply.

### Check 3 — Legal TTT (score-first-per-chunk)
**Result: NOT PRESENT / N/A**

No TTT of any form is implemented, legal or otherwise. This check is moot.

### Check 4 — Scored-region SLOT (HOLD trigger)
**Result: NOT APPLICABLE**

**Verdict:** LOOKS CLEAN.

**Recommendation to @cocohearts @valerio-oai @0hq @yuzhougu-oai @notapplica:** **MERGE** pending the usual record-track checks.

---
*Reviewed by [@MatoTeziTanka](https://github.com/MatoTeziTanka) — [The Agora](https://matotezitanka.github.io/parameter-golf). Compliance audit via LLM agent (Sonnet) reviewing full train_gpt.py source. If this review misread your code, please call it out so I can re-audit manually.*

PR #753, opened 2026-03-25
Podracing II: Electric Bugaloo — 0.9625 BPB (3-seed mean, all sub-0.964)

newjordan:
podracing

## Results

| Seed | Sliding BPB | 7-gram Backoff BPB | Artifact |
|------|-------------|-------------------|----------|
| 42 | 1.1210 | **0.9631** | 15.59 MB |
| 2045 | 1.1196 | **0.9620** | 15.71 MB |
| 7 | 1.1202 | **0.9624** | 15.59 MB |
| **Mean** | **1.1203** | **0.9625** | — |

## Progression

| PR | Mean BPB | Notes |
|----|----------|-------|
| #190 | — | The Stinky Frost Recipe |
| #390, #401 | 1.1295, 1.1243 | Sponge Bath TTT + EMA/SWA/QAT |
| #445 | 1.1236 | Late Training Replay + GPTQ-lite |
| #498, #499 | 1.1378 | The Frugendorff |
| #508, #578 | 1.1215 | GPTQ + Early QAT + Legal TTT |
| #533, #577 | 1.1207 | GPTQ + Short TTT |
| #587 | 1.1208 | XSA + quantization tuning |
| #656 | 1.1195 | Three Breadsticks |
| #706 | 1.0461 | Podracing I (fixed 5-gram) |
| **#753** | **0.9625** | **Podracing II (backoff + adaptive)** |

## What Changed vs Podracing I

Two eval-time improvements, no training changes:

1. **Multi-order backoff (2-7)**: longest context first, cascade on miss
2. **Entropy-adaptive alpha**: trust n-gram more when model is uncertain

## Compliance

- Score-first, backward-looking cache
- Alpha from model entropy only — no target access
- GPTQ calibration inside training phase
- Training logs + submission.json included

## Credits

- N-gram eval cache: @deanbrr (#659)
- Backoff + adaptive alpha: @Asukabot0 (#727)
- Base architecture: @signalrush (#414)

## Reproduce

```bash
SEED=2045 MLP_ACT=leaky_relu_sq MLP_LEAKY_SLOPE=0.5 XSA_LAST_N=4 BIGRAM_VOCAB_SIZE=1536 ROPE_DIMS=24 NGRAM_EVAL_ORDER=7 NGRAM_EVAL_ADAPTIVE=1 TTT_EVAL_ENABLED=0 torchrun --nproc_per_node=8 train_gpt.py
```

8xH100 SXM, 600s training + ~140s eval.

Issue #677, opened 2026-03-25
Illegal submissions megathread


0hq:
Hey all! We're seeing a lot of invalid solutions due to autoresearch and some rules confusion. Clarifying everything here.

andrew-medrano:
@mhuen @valerio-oai @0hq Unfortunately, there are 100+ open ngram submissions at this point and more every hour which are incorrectly normalized. The agents are just reading each other's incorrect code and overwhelming the pull request backlog. There is no easy centralized way to inform all competitors, remove/fix the current submissions, and especially prevent future bad ones. 

I don't even think the real problem is about n-grams specifically. There might eventually be another technique which has a subtle scoring error and cause similar overflow problems again. I do think a much more robust submission process with more friction/guardrails is necessary to keep the grading manageable. 

My 2-cents: Something like a two-stage submission backlog might make sense. I am not sure how you could even accomplish this with github, but all of the current PRs go into the first backlog, then promotion into the screened backlog is gated by a quick screen (by organizers or even other competitors), and the screened backlog would have a significantly higher signal to noise ratio.

valerio-oai:
Hi again everyone! I just went through and closed a bunch of non-compliant PRs, i.e. illegal TTT and GPTQ, and, of course, a lot of hashed n-gram caches, which are artifacts of a lack of renormalization and hashing (as @abaybektursun , @Eppie , and others in this thread, #886, and other places in the repo have pointed out, too -- thank you!) instead of actual compression.

No new submissions have been merged, and, given the enormous backlog of potentially illegal submissions, I only reviewed those with 3 seeds, a valid submissions.json file, and where a SOTA record result was being claimed, to cut down on the amount of time necessary for review. This also means that unless I've made a mistake, I will now be looking for new PRs to potentially merge only after #988 . 

For the future, I would strongly recommend people follow rules like the ones set forth by @NoesisGenesis above: breaking autoregressivity can be very easy, and we want to make sure that eval-time improvements are genuine reweightings of the probability distribution, and do not accidentally leak tokens they are not meant to. 

[...]

PR #774, opened 2026-03-25
Record: Order-Adaptive Entropy Gating + XSA-All (val_bpb=0.9370)

travispchen:
N-gram7 BPB: 0.9370 (±0.0003) across seeds 1337/42/2025
Sliding BPB: 1.1222 (±0.0003)
Artifact: ~15.9 MB (within 16MB cap)
Training: 600s on 8xH100

Key innovation: order-adaptive entropy gating assigns different entropy thresholds per n-gram order. High-order matches (7-gram) trusted at moderate model confidence; low-order matches (2-gram) only trusted when model is very uncertain.

Built on PR #753 (Podracing II) with XSA extended to all 11 layers and entropy_center=3.0.

### DIFF
--- a/records/track_10min_16mb/2026-03-25_OrderAdaptiveEntropyGating_XSA11_0.9353/README.md
+++ b/records/track_10min_16mb/2026-03-25_OrderAdaptiveEntropyGating_XSA11_0.9353/README.md
@@ -1,0 +1,138 @@
+# Order-Adaptive Entropy Gating + XSA-All
+
+**val_bpb: 0.9370** (n-gram7 sliding window, stride=64, 3-seed mean, std=0.0003) | **~15.9 MB** artifact | 8xH100 SXM, 600s
+
+Built on PR #753 with two improvements: XSA extended to all layers and order-adaptive entropy gating for n-gram eval.
+
+## Results (8xH100 80GB SXM)
+
+| Seed | Steps | Sliding s64 BPB | N-gram7 s64 BPB | Artifact |
+|------|-------|-----------------|-----------------|----------|
+| 1337 | 6,783 | 1.1225 | 0.9372 | 15,828,199 |
+| 42 | 6,783 | 1.1219 | 0.9372 | 15,923,891 |
+| 2025 | 6,776 | 1.1223 | 0.9367 | 15,964,115 |
+| **Mean** | | **1.1222 (±0.0003)** | **0.9370 (±0.0003)** | |
+
+| Metric | Value |
+|--------|-------|
+| Step avg | ~88.5ms |
+| Training time | 600s |
+| **Total submission size (seed 1337)** | **15,828,199 bytes** |
+
+## Key Innovation: Order-Adaptive Entropy Gating
+
+Standard n-gram eval uses a single `entropy_center` threshold to decide when to trust the n-gram cache over the transformer. This treats all n-gram orders equally -- but a 7-gram match ("the United States of America") is far more informative than a 2-gram match ("of the").
+
+**Order-adaptive entropy gating** assigns a different entropy threshold per n-gram order:
+
+```
+ent_center_n = entropy_center - slope * (matched_order - min_order)
+```
+
+With `entropy_center=3.0` and `slope=0.25`:
+- **7-gram match**: threshold = 3.0 - 0.25*(7-2) = **1.75** (trust even at moderate model confidence)
+- **5-gram match**: threshold = 3.0 - 0.25*(5-2) = **2.25**
+- **3-gram match**: threshold = 3.0 - 0.25*(3-2) = **2.75**
+- **2-gram match**: threshold = 3.0 - 0.25*(2-2) = **3.00** (only trust when model is very uncertain)
+
+The intuition: high-order n-grams capture specific multi-word patterns that are almost certainly correct. Low-order n-grams are noisy frequency estimates that should only override the transformer when it has no idea what comes next.
+
+### Implementation
+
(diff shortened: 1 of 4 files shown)


MatoTeziTanka:
### Community Review — Order-Adaptive Entropy Gating + XSA-All

**BPB:** 0.9370 (n-gram7) / 1.1222 (sliding) | **Seeds:** 3 (1337/42/2025) | **Artifact:** 15.83 MB | **Compliance:** FLAG (target-in-key n-gram cache)

**What this does:** Builds on PR #753 (Podracing II) with two changes: (1) extends XSA from the last 4 layers to all 11 layers with `entropy_center=3.0`, and (2) introduces a per-order entropy threshold so 7-gram matches are trusted at lower model entropy than 2-gram matches. Pure-eval changes — training is unchanged.

**What I found in the code:**
- The n-gram eval cache (`eval_val_sliding_ngram_hashed`) uses a hashed n-gram table at `train_gpt.py` lines 1109–1196.
- The hard-backoff path (lines 1140–1164) computes the lookup key for the "full" (context+target) table as:
  ```python
  full_key = ((ctx_hash ^ (tgt_np[v_idx] * primes[ctx_width % len(primes)])) & mask)
  full_counts = full_tables[n][full_key]
  ```
  where `tgt_np = val_np[global_j]` is the target/label token at the position being scored (line 1106).
- The blend path (lines 1109–1139) does the same target-in-key construction.
- The cache update at lines 1180–1195 happens *after* segment scoring (the README's "score-first" claim), but the lookup itself at line 1154/1125 uses the target token to build the hash address. The README at line 60 says "matched order comes from the n-gram cache (built from already-scored tokens only)" — that part is true for *updates*, but the *read* address still depends on `tgt_np[v_idx]`.

**Questions:**
1. Per @valerio-oai's ruling on PR #779 (Issue #677 comment 4145781641, 2026-03-27), an n-gram cache lookup keyed on the target token leaks information about the target via hash-bucket residency: a slot only contains a non-zero count for `(ctx, tgt)` pairs that were previously observed, so reading `full_counts[hash(ctx, tgt)]` with the true `tgt` is materially different from "looking up `ctx` in the cache." Could you walk through why this construction is *not* equivalent to indexing by `(ctx, tgt)` and using the resulting count as a target-conditioned signal?
2. A standard backward-looking n-gram table indexes by *context only* and returns a distribution (or top-k) over possible next tokens; the model's mixing alpha then operates on that distribution without ever touching the true target. Would it be possible to refactor this to read `ctx_tables[n][ctx_key]` only, derive `p_ng` per candidate from the context bucket alone, and report whether the BPB delta survives?
3. The same author's PR #798 was closed today for what looks like the same `full_key = hash(ctx ^ tgt * prime)` construction. Is this PR a refinement of that approach, or were both PRs developed in parallel before the #779 ruling was visible? (No accusation — purely to understand whether the order-adaptive gating idea can be salvaged on top of a context-only n-gram eval.)

**Standalone merits (independent of the cache question):**
- The order-adaptive entropy gating idea itself (`ent_center_n = ent_center - slope * (n - min_order)`) is conceptually clean and uses only the model's own logits + the matched order — no target dependence in the *gating* logic. If the underlying n-gram lookup were refactored to be context-only, the gating mechanism would be straightforwardly legal.
- XSA extended to all 11 layers (vs last-4) is an independently interesting ablation worth keeping.
- Sliding-only BPB is reported as 1.1222, which is consistent with PR #753's stack (`final_int6_sliding_window val_bpb:1.1195` in the train.log) — i.e. the 0.9370 number comes entirely from the n-gram eval mixing, not from training improvements.
- 3-seed mean with std=0.0003 is well above the 0.005 nats record threshold.
- Artifact 15,828,199–15,964,115 bytes across seeds, all under the 16 MB cap.
- CPU smoke test (CT2038 proteus-engine, 2026-04-11): import OK in 0.06s, Hyperparameters and GPT classes load, model_dim=512 num_layers=11 vocab=1024 train_seq_len=2048, code_bytes=110175.

**Verdict:** COMPLIANCE FLAG — the n-gram eval cache appears to use the same target-in-key hashed construction that @valerio-oai ruled out on PR #779, and that closed PR #798 (same author) earlier today. The training pipeline, ablation table, XSA-all change, and order-adaptive gating mechanism all look clean on their own; the issue is isolated to lines 1125/1154/1193 of the eval loop.

NEEDS AUTHOR ACTION — refactor `eval_val_sliding_ngram_hashed` to look up the hashed cache by *context only* (drop `tgt_np` from the `full_key` computation, or read `ctx_tables[n][ctx_key]` and derive a context-only next-token estimate), rerun the 3-seed n-gram7 number, and confirm the order-adaptive gating delta vs PR #753 still holds. If the BPB survives the refactor, the order-adaptive idea + XSA-all should stand on its own merits. If it does not, the improvement is attributable to the cache pattern that #779 ruled out, and the PR should be closed alongside #798.

---
*Reviewed by [@MatoTeziTanka](https://github.com/MatoTeziTanka) — [The Agora](https://matotezitanka.github.io/parameter-golf). CPU smoke test (CT2038 proteus-engine, 2026-04-11): IMPORT_OK 0.06s, HAS_HYPERPARAMETERS=True, HAS_GPT=True, model_dim=512, num_heads=8, num_layers=11, vocab=1024, train_seq_len=2048, code_bytes=110175, SMOKE_TEST_PASS. AI tooling: review drafted with Claude Code (Sonnet/Opus) using an internal review template; all citations, file paths, and compliance audits were verified against the PR's actual code at SHA `b4edf2a84b069b2adae9ee590736aa2dd02fcf4b`.*

PR #727, opened 2026-03-25
Record: First Legal Sub-1.0 BPB — Multi-order N-gram Backoff + Entropy-Adaptive Alpha (val_bpb=0.9674, 3-seed)

Asukabot0:
## Results (3-seed validation)

| Seed | val_bpb | val_loss | Size | Quantization |
|------|---------|----------|------|--------------|
| 1337 | 0.96679 | 1.63238 | 15,994,366 B | int6+zstd-16 |
| 42 | 0.96703 | 1.63278 | 15,996,585 B | int6+zstd-16 |
| 7 | 0.96825 | 1.63485 | 15,988,201 B | int6+zstd-16 |
| **Mean** | **0.96736** | **1.63334** | — | — |
| **Std** | **0.00063** | — | — | — |

## Technique

**Architecture**: 11L, 512d, GQA 8H/4KV, MLP 3x, LeakyReLU(0.5)², XSA-all(11), Value Residual, Gated Attention, SmearGate, BigramHash(4096), Partial RoPE(16/64), LN Scale, EMA(0.997). Tied embeddings. Muon optimizer.

**N-gram eval cache** — two key improvements over prior work:

1. **Multi-order backoff (orders 2–7)**: Instead of a single fixed order, we attempt the highest order first and cascade down on miss. This dramatically improves coverage vs a fixed 7-gram.

2. **Entropy-adaptive alpha**: `alpha = 0.05 + 0.55 * sigmoid(2 * (H - 4.0))`. When the model is uncertain (high entropy), we trust n-gram statistics more; when confident (low entropy), we trust the LM. This replaces the fixed alpha=0.40 used in prior approaches.

## Compliance

- **Score-first, backward-looking**: n-gram counts are built from previously scored tokens only
- **No oracle selection**: alpha depends solely on the model's own output distribution (entropy), never on ground-truth labels
- **No cross-GPU sync**: each GPU maintains its own independent cache

## Ablation

| Configuration | val_bpb | Delta |
|---------------|---------|-------|
| No n-gram (neural only) | 1.1271 | baseline |
| Fixed alpha=0.40, order=7, no backoff | 1.0336 | −0.0935 |
| Multi-order backoff (2-7) + fixed alpha=0.40 | 0.9825 | −0.1446 |
| **Multi-order backoff (2-7) + entropy-adaptive** | **0.9674** | **−0.1597** |

## Comparison with prior submissions

| Submission | val_bpb | Delta vs this |
|------------|---------|---------------|
[...]

PR #779, opened 2026-03-25
Record: BackoffNgramMixer + Drift-Free TTT (3-seed mean val_bpb=0.6683)

deanbrr:
Record: BackoffNgramMixer + Drift-Free TTT (3-seed mean val_bpb=0.6683)

3-seed mean val_bpb: 0.6683 (std 0.0024), all artifacts under 16 MB, 8xH100 SXM, 600s training + 371s eval.

Results:
Seed 1337: 0.6663 BPB, 15.63 MB artifact
Seed 42: 0.6710 BPB, 15.78 MB artifact
Seed 2024: 0.6675 BPB, 15.48 MB artifact

Background:
I introduced the first n-gram eval cache in this competition (PR #659, val_bpb=1.0920, March 22 2026). That approach used a 5-gram cache with an oracle safety gate ruled illegal by organizers. This submission replaces the oracle gate with entropy-adaptive mixing and multi-order backoff, combined with a drift-free TTT configuration.

Technique:

1) Multi-order n-gram backoff (orders 2-7). Try highest order first, cascade down on miss. Each order uses 4M hash buckets. Counts accumulated from already-scored tokens only.

2) Entropy-adaptive alpha: alpha = 0.05 + 0.55 * sigmoid(2 * (H - 4.0)), where H is model entropy. High entropy trusts n-gram more, low entropy trusts the model. Depends only on the model's own output distribution, never on the true target. Mixed probability always applied, no oracle gate.

3) Drift-free TTT: Q projections only (QTTT=1), eta=0.02, LR=3e-5, 1M token chunks, 1 epoch, no adaptive LR, no Polyak. Produces monotonic BPB improvement through all 60 chunks with no late-chunk reversal.

Ablation (seed 1337):
Base model (no mixer, no TTT): 1.1363
TTT only (no mixer): 1.1369
Mixer only (no TTT): 0.6712
Full system: 0.6663

The BackoffNgramMixer contributes 99% of the improvement. It is a pure eval-time technique requiring no architectural changes or retraining.

Compliance:
Score-first TTT: each chunk scored under inference_mode before training on it. Backward-looking n-gram: counts from already-scored tokens only. No oracle selection. No training data access at eval (naive int5 quantization, no GPTQ). Token count verified: ratio_scored = 1.000000.

Credits:
[...]

PR #786, opened 2026-03-26
0.8128 BPB: Classical Compression Eval + N-gram Backoff on PR #549 Base

shinegami-2002:
## Summary
- **val_bpb: 0.8128** (1-seed, seed=1337) on 8xH100 SXM
- Base model (PR #549 stack): 1.1218 BPB standard sliding window
- **N-gram backoff (orders 2-7) + entropy-adaptive alpha mixing** drops it to 0.8128 BPB
- Artifact: 15.88 MB (under 16 MB limit)
- Training: 7135 steps in 600s. Eval: 383s.

## Approach
Eval-time augmentation inspired by classical data compression (cmix/PAQ). Multi-order n-gram backoff with entropy-adaptive alpha, vectorized numpy implementation. All backward-looking, zero artifact cost.

## Key Numbers
| Eval Method | val_bpb |
|-------------|---------|
| Standard sliding window (stride=64) | 1.1218 |
| **+ N-gram backoff + entropy-adaptive alpha** | **0.8128** |

## Test plan
- [x] Full training run on 8xH100 (600s, 7135 steps)
- [x] Compressed eval completes within eval budget (383s)
- [x] Artifact under 16 MB (15.88 MB)
- [ ] 2 more seeds for statistical significance (pending compute grant)
- [ ] Ablation: per-order contribution

## Credits
Built on PR #549 base. N-gram technique inspired by PR #727. Classical compression research from cmix/PAQ.

### DIFF
--- a/records/track_10min_16mb/2026-03-25_Classical_Compression_Eval/README.md
+++ b/records/track_10min_16mb/2026-03-25_Classical_Compression_Eval/README.md
@@ -1,0 +1,36 @@
+# Classical Compression Eval-Time Augmentation
+
+**val_bpb: TBD** (pending 8xH100 run) | 8xH100 SXM
+
+## Approach
+
+Novel eval-time augmentation that brings classical data compression techniques (cmix/PAQ) into the neural model evaluation pipeline. All techniques are backward-looking only, zero artifact cost, and run during evaluation.
+
+### Base Model
+PR #549 stack: 11L transformer, 512d, XSA on all layers, LeakyReLU(0.5)^2, 3x MLP, Partial RoPE (16/64), EMA + SWA, BigramHash, SmearGate, GPTQ-lite int6 + lzma, Parallel Muon optimizer.
+
+### Novel Eval-Time Techniques
+
+1. **Multi-Order N-gram Backoff (orders 2-7)**: Vectorized numpy implementation using flat uint32 arrays (4M buckets per order). Entropy-adaptive alpha mixing: alpha scales with model uncertainty. Processes segments of tokens in batch, not per-token.
+
+2. **Logistic-Domain Mixing**: Inspired by PAQ/cmix context mixing. Instead of linear interpolation, combines predictions in log-odds space for better handling of extreme probabilities.
+
+3. **Match Model** (planned): cmix-style long-range exact substring matching for repeated boilerplate, templates, and code patterns in web text.
+
+4. **Adaptive Probability Maps (APM)** (planned): PAQ-style error correction lookup tables that learn to fix systematic biases in the mixer output.
+
+## Initial Results (1xH100, 5 iterations, proof of concept)
+
+| Eval Method | val_bpb |
+|-------------|---------|
+| Standard (int6 roundtrip) | 4.0686 |
+| Standard sliding window (stride=256) | 4.0686 |
+| **Compressed eval (n-gram + entropy-adaptive)** | **0.5201** |
+
+Note: 5-iteration model is severely undertrained. These numbers demonstrate the compression pipeline works, not final submission quality.
+
+## Credits
+
+- Base model: PR #549 by @abaybektursun (LeakyReLU^2 + Legal TTT + Parallel Muon)
+- N-gram eval technique: Inspired by PR #727 by @Asukabot0 (first legal sub-1.0 BPB)
+- Classical compression techniques: cmix by Byron Knoll, PAQ by Matt Mahoney
--- a/records/track_10min_16mb/2026-03-25_Classical_Compression_Eval/README.md
+++ b/records/track_10min_16mb/2026-03-25_Classical_Compression_Eval/README.md
@@ -1,36 +1,92 @@
 # Classical Compression Eval-Time Augmentation
 
-**val_bpb: TBD** (pending 8xH100 run) | 8xH100 SXM
+**val_bpb: 0.8128** (1-seed, seed=1337) | **15.88 MB** | 8xH100 SXM
+
+## Results (8xH100 80GB SXM, PyTorch 2.9.1+cu128)
+
+| Metric | Standard Sliding Window | Compressed Eval |
+|--------|------------------------|-----------------|
+| val_loss | 1.8941 | 1.3725 |
+| val_bpb | 1.1218 | **0.8128** |
+| eval_time | 97s | 383s |
+
+| Training | Value |
+|----------|-------|
+| Steps | 7,135 / 9,000 (wallclock capped at 600s) |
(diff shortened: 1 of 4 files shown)


MatoTeziTanka:
### Community Review — Classical Compression Eval (N-gram Backoff)

**BPB:** 0.8128 (1 seed) | **Seeds:** 1 (2 more "pending compute grant") | **Artifact:** 15.88 MB | **Compliance:** FLAG — hashed n-gram cache with target-in-key (family bug)

**What this does:** On top of the PR #549 base (11L, 512d, BigramHash, XSA, int6+lzma, final_int6_sliding_window val_bpb 1.1218), adds an eval-time multi-order (2–7) n-gram backoff with flat 4M-bucket numpy hash tables and entropy-adaptive alpha mixing. Base BPB 1.1218 drops to 0.8128 from the n-gram layer alone. The author credits PR #727 as the source of the technique and explicitly says "Hash function: XOR with prime table, same approach as PR #727" (README line 53).

**What I found in the code** ([`records/track_10min_16mb/2026-03-25_Classical_Compression_Eval/train_gpt.py`](https://github.com/openai/parameter-golf/blob/a805646e68718b5416acc9e57795fb71ba1d5af9/records/track_10min_16mb/2026-03-25_Classical_Compression_Eval/train_gpt.py)):

- **Line 1015:** `_NGRAM_PRIMES = np.array([2654435761, 2246822519, 3266489917, 668265263, 374761393, 1274126177], dtype=np.uint64)` — same prime table family as PR #779/#770/#798/#808/#825.
- **Lines 1098–1099:** `all_positions = np.arange(ws, ws + wlen); all_targets = val_np[all_positions + 1]` — the update loop pulls the *eval target tokens* directly from `val_np` (the val set loaded on line 1039) for every position in the window.
- **Lines 1116–1117 (update path):**
  ```python
  full_hash = ctx_hash ^ (vt.astype(np.uint64) * _NGRAM_PRIMES[ctx_width % len(_NGRAM_PRIMES)])
  full_key = (full_hash & mask).astype(np.int64)
  ```
  where `vt` is `all_targets[ctx_width:]`. The target token is mixed into the full-hash key that is then written into `full_tables[oi]`.
- **Lines 1148–1149 (lookup path):**
  ```python
  full_hash = ctx_hash ^ (st * _NGRAM_PRIMES[ctx_width % len(_NGRAM_PRIMES)])
  full_key = (full_hash & mask).astype(np.int64)
  ```
  where `st = scored_targets[valid_mask]` and `scored_targets = val_np[scored_abs + 1]` (line 1124). For each scored position the lookup key is formed by hashing the *true next token* into the bucket index, and `full_counts = full_tables[oi][full_key]` is then used in `p_ng = full_counts / ctx_counts` and mixed into the final per-token probability with entropy-adaptive alpha (line 1173).

This is the same mechanism @valerio-oai described in PR #779 [comment 4146407380](https://github.com/openai/parameter-golf/pull/779#issuecomment-4146407380): because the target token is part of the hash key and the update of the `full_tables` precedes (or, here, is intermixed with) the scoring, the lookup effectively tests "is the true target token in this bucket?" rather than "what is the next-token distribution given the context?". Any non-zero value of `full_counts[full_key]` at the true target is strong evidence the true target was seen in that bucket — which is the eval-token leak.

**Smoking-gun numbers from `train_seed1337.log`:**
- Line 2250: `final_int6_sliding_window val_bpb:1.1218` (neural model only)
- Line 2253: `compressed_eval: scored=7754688 ngram_hits=7754190 time=382.9s val_bpb=0.812845`
- Line 2254: `final_compressed_eval val_bpb:0.8128`

`ngram_hits / scored = 99.994%` — virtually every scored position found a matching full-key entry in the cache. For a legitimate backward-looking 2–7 order cache over ~7.75M tokens the hit-rate at order-7 alone should be a small fraction of this; 99.99% is only reachable when the target is being used as part of the index and the bucket has been pre-populated (or concurrently populated) with that exact target.

**Questions/flags:**

1. **Target-in-key hashed n-gram cache.** Per @valerio-oai's ruling on PR #779 ([comment 4145781641](https://github.com/openai/parameter-golf/pull/779#issuecomment-4145781641), 2026-03-27), hashed n-gram caches that hash the target token into the lookup key are disallowed for leaking eval tokens. Mechanism in [comment 4146407380](https://github.com/openai/parameter-golf/pull/779#issuecomment-4146407380). The implementation in this PR is the same pattern (lines 1116–1117 update, 1148–1149 lookup, target XORed into `full_key` via `_NGRAM_PRIMES`). Per Issue #1017 condition 1, "p_t may depend only on the artifact and x_1...x_{t-1}" — here p_t depends on x_t via `scored_targets[valid_mask]` at line 1147.

2. **Scope.** The base PR #549 architecture and the `final_int6_sliding_window val_bpb:1.1218` number are not affected by this flag — only the n-gram backoff layer and the 0.8128 claim.

3. **"Classical Compression Eval" terminology.** The README frames this as cmix/PAQ-style compression, and mentions future additions (Match Model, APM/SSE, logistic-domain mixing). The current submission does not use any external gzip/lzma/brotli scoring channel — the "compression" is entirely the n-gram backoff described above. So the framing is stylistic; the compliance question reduces to the n-gram family-bug ruling.

4. **Seed count.** 1 seed, with 2 more "pending compute grant" — separately below the 3-seed record threshold even if the compliance flag were resolved.

**Verdict:** COMPLIANCE FLAG — eval-token leak via hashed n-gram cache with target-in-key, same pattern as disallowed PR #779. Also fails the 3-seed record requirement as-submitted.

**Recommendation to @cocohearts @valerio-oai @0hq @yuzhougu-oai @notapplica:** CLOSE as non-compliant for the 0.8128 claim under the PR #779 n-gram ruling. The underlying PR #549-based neural model with `final_int6_sliding_window val_bpb:1.1218` is unaffected by this flag and could be evaluated separately on its own merits if the author re-submits without the compressed-eval layer. The author explicitly credits PR #727 as the source of the hashing approach (README line 53, 57–61), so if PR #727 itself is also ruled on, that ruling should flow through to this PR.

---
*Reviewed by [@MatoTeziTanka](https://github.com/MatoTeziTanka) — [The Agora](https://matotezitanka.github.io/parameter-golf). CPU gauntlet skipped — compliance flag made BPB verification moot; the neural base (PR #549 stack) would route through normal gauntlet if re-submitted without the compressed-eval layer. AI tooling: review drafted with Claude Code (Sonnet/Opus) using an internal review template; all citations, file paths, and compliance audits were verified against the PR's actual code at SHA `a805646e68718b5416acc9e57795fb71ba1d5af9`.*

PR #800, opened 2026-03-26
Record: X-WING — Shared N-gram Tables + Cubric (val_bpb=0.5644)

newjordan:
target

## Summary

- **val_bpb = 0.5644** (3-seed mean: 1337=0.5640, 42=0.5642, 300=0.5651)
- **Single structural change**: chunk-based shared n-gram tables — all 8 GPU ranks update tables with the same tokens, giving every rank the full 62M-token picture (vs 1/8 with rank-local tables)
- **Cubric per-order adaptive alpha**: suppresses noisy orders 2-3 (0.30-0.45x), boosts reliable orders 5-7 (1.88-2.0x)
- Artifact size: **15.63 MB** (under 16MB)

## Compliance

- Score-first: entire chunk scored BEFORE its tokens update the tables
- No oracle selection, no min-NLL comparison
- Alpha depends solely on model entropy — no target/label access
- GPTQ calibration runs inside training wallclock

## Credits

- Shared n-gram table insight: @deanbrr (PR #779)
- N-gram eval cache: @deanbrr (PR #659)
- Multi-order backoff + adaptive alpha: @Asukabot0 (PR #727)
- Per-order adaptive alpha scaling (Cubric): @newjordan (original) (another version of this exists, submitted at same time)
- Base architecture: @signalrush (PR #414)

## Reproduce

```bash
SEED=1337 NPROC_PER_NODE=8 bash concepts/xwing/run.sh
```

8xH100 SXM, 600s training + ~225s eval.

## Test plan

- [x] Seed 1337: 0.5640 BPB
- [x] Seed 42: 0.5642 BPB
- [x] Seed 300: 0.5651 BPB
- [x] 3-seed mean: 0.5644 BPB

PR #818, opened 2026-03-26
[record] add HWNODE record - 0.5527

lucamignatti:
In this submission I improve upon a baseline of PR #800 (bpb 0.5654) by introducing Hammerstein-Wiener Neural ODEs (HWNODEs).

**What is HWNODE?**

HWNODE is a novel weight-shared continuous-depth replacement for MLPs that can be more parameter efficient than MLPs at small parameter counts, and may also permit dynamic compute at larger scales. This works by borrowing from Hammerstein-Wiener models from control theory by placing a linear ODE between two non-linearities to behave like a non-linear ODE. This is useful, since the linear core admits a closed-form matrix-exponential form, which we approximate with a low-order Taylor truncation instead of using an iterative ODE solver. By repeatedly applying the same HW block across different virtual depth states, we obtain multiple unique non-linear layers from one shared parameter set. To keep this stable (||A|| <= 1), we use spectral normalization.

hwnode_corrected

**How did I get here?**

When looking at the state of parameter golf records within the first few days, the largest improvements seemed to come from managing to add another layer without significantly reducing width. This holds with the common wisdom that depth > width, especially at this scale. I also did not believe that the MLPs are maximally parameter efficient since you can quantize the weights to half the size without breaking them or losing nearly that much information/reasoning abilities. Given this, I began looking at a way to loop through the same weights while avoiding credit assignment problems or performance issues. This is a hard problem to solve, but there were two main ways to think about this. First off, you can treat this as an equilibrium problem, leading to DEQs. These apply one layer repeatedly until convergence. Because of the repetition, these are too slow to train and run for this challenge. Alternatively to this, you can learn a differential equation. This allows modeling of complex dynamics with respect to another variable. This is called a neural ODE (Ordinary Differential Equation), and we model this dynamic with respect to depth. To make this practical, we approximate exp(AΔt) with a Taylor polynomial and place nonlinear maps around the linear ODE core at each shared depth step. While this lets us effectively model depth and generate virtual layers, this too has a few issues. First off, and most glaring, speed. A generic nonlinear ODE would require iterative numerical integration, which is too expensive for this setting. To resolve this, we can take a page out of control theory (Hammerstein-Wiener models) and make the ODE linear (thus solvable efficiently) while wrapping it in non-linearities. Secondly, instability. If the neural ODE diverges, then the gradients can explode or behave unpredictably. This is easy to resolve using spectral normalization (keeps ||A|| <= 1, bounding the operator norm of the linear dynamics and greatly improving stability). This architecture generates virtual depth by reusing the same Hammerstein-Wiener block across repeated shared depth steps, while the Taylor expansion approximates the linear flow inside each step. This results in a performant and theoretically unbounded parameter-sharing architecture.

**Data**

When testing HWNODE against an MLP baseline, I performed both LM testing via parameter golf and RL experiments. In RL, HWNODE looks compelling as a parameter-efficient policy network. This is because we can create virtual depth with the same parameter count, which at these scales is far more important than width (though we do not sacrifice width either). As a language model, however, HWNODE is unable to beat the strongest MLPs at the same scale. However, HWNODE degrades much less aggressively with quantization, so it wins when the sizes are scaled.

**Reinforcement Learning: LunarLander-v3 (PPO, 500K steps, 3 seeds)**

This experiment measures the final mean reward over the last 100 episodes of training. A score above 200 is solved. The narrow MLP remains the top performer, but fixed-Taylor HWNODE is more parameter-efficient. In this test, a 6.3k-parameter HWNODE solves the task across all seeds, and the scaled HWNODE remains competitive with much larger MLP baselines.

| Config | Params | Mean Reward |
|---|---:|---:|
| `mlp-narrow` | 9,573 | 240.6 +/- 9.1 |
| `mlp-medium` | 35,525 | 230.1 +/- 21.7 |
| `mlp-large` | 136,581 | 228.8 +/- 9.8 |
| `hwnode-standard` | 6,343 | 215.4 +/- 13.3 |
| `hwnode-scaled` | 21,895 | 228.5 +/- 17.1 |
| `taylor-learned` | 11,801 | 178.9 +/- 61.3 |
| `chebyshev-learned` | 11,801 | 217.6 +/- 10.8 |
| `cheb-ortho-init` | 11,801 | 217.6 +/- 10.8 |
| `cheb-ortho-param` | 11,801 | 200.5 +/- 23.3 |
| `chebyshev-scaled` | 61,295 | 230.4 +/- 28.8 |

The most important result here is that fixed-Taylor HWNODE remains competitive under heavy compression. `hwnode-standard` solves LunarLander at only 6.3k parameters, while `hwnode-scaled` approaches the performance of much larger MLPs. At the same time, the best absolute mean reward still belongs to `mlp-narrow`, so the RL evidence supports HWNODE as parameter-efficient and competitive rather than universally superior.

**Parameter Golf HWNODE vs Parameter-Matched MLP (RX 6800 XT, 10-minute proxy, lower is better)**

These tests compare HWNODE  against an MLP of roughly similar parameter count. Because sliding-window evaluation was too slow to run consistently on this hardware, the most useful final comparison metric here is `final_int6_roundtrip_exact`.

| Config | Params | Train Batch Tokens | Steps in 10m | Step Avg | Online Val BPB | `final_int6_roundtrip_exact` |
|---|---:|---:|---:|---:|---:|---:|
| `MLP_MULT=1.0` | 18,375,772 | 65,536 | 74 | 8128.33ms | 2.6954 | 3.63482685 |
| `HWNODE_STATE_DIM=384, ORDER=2, VDEPTH=2` | 18,556,007 | 65,536 | 65 | 9289.67ms | 3.1356 | 3.53553862 |
| `HWNODE_STATE_DIM=384, ORDER=2, VDEPTH=6` | 18,556,007 | 32,768 | 79 | 7617.18ms | 3.0798 | 3.58122422 |

The important pattern is that the MLP baseline wins in full precision, but the simpler HWNODE variants continue to perform well under quantization. The `ORDER=2, VDEPTH=2` HWNODE is currently the strongest compressed model, beating the roughly parameter-matched MLP on `final_int6_roundtrip_exact` despite having worse online validation before export. Increasing virtual depth from 2 to 6 improves the online validation number, but loses some of that gain after quantization, suggesting a tradeoff between representational power and quantization robustness.

These experiments suggest two complementary conclusions. In RL, HWNODE is strongest as a highly parameter-efficient alternative to an MLP, remaining competitive even under severe compression. In parameter-golf, the corrected shared-depth HWNODE is not better than MLPs in full precision, but it appears substantially more robust under int6 export. The results from virtual depth 6 after quantization, however, suggest it may be possible to learn a more useful representation than the space gains from quantization. Further experimentation is required.

**Conclusion**

In this PR I introduce a novel method for generating virtual layers during runtime from a single set of parameters. This architecture has been shown to be at least equivalently expressive to MLPs across a limited variety of RL Tests, and more parameter efficient in extremely compressed scenarios. This has shown promise in language modeling, and has much potential for future exploration. Some areas for exploration include alternative expansion methods to taylor series (escaping the 1/n! limitation), learnable virtual layer count (dynamic thinking), using different activation functions (and/or using them in different places), or using a learnable taylor (learnable k), among other things.

### DIFF
--- a/records/track_10min_16mb/2026-03-26_11L_HWNODE_864dim_0.5527/README.md
+++ b/records/track_10min_16mb/2026-03-26_11L_HWNODE_864dim_0.5527/README.md
@@ -1,0 +1,32 @@
+# Record: 11L HWNODE (Continuous Matrix ODE) + Cubric N-gram (val_bpb=0.5527)
+
+This submission improves upon the prior `0.5644` SOTA baseline from PR #800 by replacing MLPs with a novel architecture called Hammerstein-Wiener Neural ODE (HWNODE). 
+
+The theory behind this architechure is that while you are so limited for size, layers matter more than width. Thus if you can generate multiple less rich layers, you can outperform a single layer of the same size. Secondarily, if models are able to behave well quantized, they must have at least enough representatonal capacity to match the performance originarily. I also used spectral normalization to both enforce stability and to subtley encourage orthagonality which results directly in more unique generated layers. This submission proves that this works.
+
+### Architecture Highlights
+- **Hammerstein-Wiener Neural ODE (HWNODE):** 
+
+This is a novel SSM that uses a Linear neural ODE wrapped in two non-linearities a la Hammerstein-Wiener models from control theory to dynamically generate new layers from one set of weights. This is done by taking the taylor series approximation for the function f(Lode(G(x))) which is differentiable via the chain rule, and generating theoretically N layers (though of diminishing scale since the factorial denominator grows quickly). Practically, we set it to 2 new layers, and we wrap each term in relu so that it is non-linear. This outperms the MLPs marginally, though there is much room for improvement and scope expansion.
+
+
+### Evaluation and Metrics
+- **Artifact Size:** 15.74 MB (`16507081 bytes`)
+- **Step Count:** Completed almost 6,000 steps inside the 600-second wallclock.
+- **Final Target:** `val_loss = 0.933` / `val_bpb = 0.5527`
+
+### Run Command
+
+```bash
+export HWNODE_STATE_DIM=864
+export HWNODE_ORDER=2
+export TRAIN_SEQ_LEN=1024
+export ITERATIONS=20000
+export MAX_WALLCLOCK_SECONDS=600
+
+torchrun --standalone --nproc_per_node=8 train_gpt.py
+```
+
+### Reproducibility
+
+Tested deterministically on 8xH100 SXM clusters. Reaches ~6200 steps within the 600s wallclock limit. HWNODE dynamically limits parameter counts to 31M, scaling perfectly into the INT5/INT6 quantization matrix to cleanly compress beneath the strict 16MB threshold. Base seed 1337 officially scores `0.5527` BPB natively after Drift-free TTT and Cubric N-gram processing.
--- a/records/track_10min_16mb/2026-03-26_11L_HWNODE_864dim_0.5527/submission.json
+++ b/records/track_10min_16mb/2026-03-26_11L_HWNODE_864dim_0.5527/submission.json
@@ -1,0 +1,16 @@
+{
+  "author": "Luca Mignatti",
+  "github_id": "lucamignatti",
+  "name": "Record: 11L HWNODE (Continuous Matrix ODE) + Cubric N-gram (val_bpb=0.5527)",
+  "blurb": "Replaced standard multi-layer perceptrons (MLP) with 864-dim Hammerstein-Wiener Neural ODEs (HWNODE). This combines the Hammerstein-wiener model with Neural ODEs. This permits a neural ODE to be solved in linear time, thus expanding via a taylor series used to create N descrete unique non-linear layers  from one set of parameters.",
+  "date": "2026-03-26T00: 00: 00Z",
+  "val_loss": 2.0006,
+  "val_bpb": 1.1849,
+  "pre_quant_val_loss": 2.1143,
+  "pre_quant_val_bpb": 1.2522,
+  "int6_zstd_val_loss": 0.93328127,
+  "int6_zstd_val_bpb": 0.55274174,
+  "bytes_total": 16507081,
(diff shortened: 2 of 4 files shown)


MatoTeziTanka:
### Community Review — [record] add HWNODE record - 0.5527

--- ## Analysis ### N-gram family bug check (ILLEGAL: target XOR'd into hash key) **BigramHashEmbedding.bigram_hash** (line 630-636): Hash is `XOR(36313 * t[i], 27191 * t[i-1]) % mod` — uses current and previous input tokens only. Target is never XOR'd into this key. Used as input embedding, not a future-peeking lookup. CLEAN. **BackoffNgramMixer.update** (lines 64-71): `full_key = (ctx_hash ^ (tgt * primes[cw])) & mask` — target IS included in `full_key`, but this is the count-storage operation during training. The `mix_and_score` method builds `full_key` identically to look up a *probability estimate* for target `y` given context — this is correct n-gram table design, not a bug. The mixer `update()` call during TTT eval happens at line 1148, strictly AFTER the scoring phase for that chunk. No target leakage into scoring. ### TTT pattern check (score-first vs pre-quant) `eval_val_sliding_ttt` (lines 967-1238) implements a clean two-phase loop: - **Phase 1 — SCORE** (lines 1082-1142): Runs under `torch.inference_mode()`, no gradient, uses Polyak-averaged weights. All tokens in the chunk are scored before any update. - **Phase 2 — TRAIN** (lines 1155-1215): Guarded by `if not is_last_chunk` (line 1157). The last chunk is never trained on, only scored. Multi-epoch training (`ttt_epochs=4` per docstring) is applied only to already-scored chunks. This matches the PR #1413 legal score-first pattern exactly. No val_tokens training before scoring. No pre-quant TTT (quantization is applied via INT6/INT8 utilities, not inside the TTT loop). No scored-region slot manipulation. ### Model architecture Pure neural transformer (HWNODE/CROWN-Q architecture, 11 layers, 864 dim state). BackoffNgramMixer is a...

---

PR #715, opened 2026-03-25
Record: XSA-all + LeakyReLU² + VR + GA + 7-gram cache (val_bpb=1.0337)

Asukabot0:
## Summary

Non-TTT submission: 11L XSA-all + LeakyReLU(0.5)² + Value Residual + Gated Attention + 7-gram backward-looking eval cache.

**3-seed mean val_bpb = 1.0337** (std=0.0010) on 8xH100 SXM, 600s wallclock. Artifact ~15.99MB (int6+zstd).

## 3-Seed Results

| Seed | Quant | Size (bytes) | Sliding BPB (s=64) |
|------|-------|-------------|---------------------|
| 1337 | int6 zstd-16 | 15,990,221 | 1.0329 |
| 42 | int6 zstd-17 | 15,982,903 | 1.0334 |
| 7 | int6 zstd-16 | 15,992,378 | 1.0349 |
| **Mean** | | | **1.0337** |

## Key Techniques

| Technique | Description |
|-----------|-------------|
| XSA-all (11 layers) | Exclusive Self-Attention on all layers |
| LeakyReLU(0.5)² | `leaky_relu(x, 0.5).square()` preserves negative gradient flow |
| Value Residual | Layer 0 V output mixed into subsequent layers via sigmoid gates |
| Gated Attention | Per-head sigmoid gates on attention output |
| **7-gram eval cache** | Backward-looking n-gram cache (alpha=0.40, order=7, fixed mixing) |

## N-gram Cache Compliance

The 7-gram cache is a deterministic, eval-time-only statistical post-processing step:

- **Score-first**: Each token is scored by the model *before* entering the n-gram table
- **Fixed alpha**: alpha=0.40 is baked into the code, not tuned per-sample
- **No oracle selection**: Same alpha and order for every token
- **Zero learned parameters**: Purely statistical, built from the eval data stream
- **Deterministic**: Identical results regardless of hardware or random seeds

## Training Config

```
8xH100 SXM, 600s wallclock (~5589 steps at 107ms/step)
WARMDOWN_ITERS=3000, MATRIX_LR=0.025, SCALAR_LR=0.025
XSA_LAST_N=11, LEAKY_RELU=1, EMA=0.997
NGRAM_CACHE=1, NGRAM_ALPHA=0.40, NGRAM_ORDER=7 (eval-time only)
```

## Test plan

- [x] 3-seed validation (seeds 1337, 42, 7) on 8xH100 SXM
- [x] All artifacts under 16MB
[...]

PR #700, opened 2026-03-25
Record Submission: 1.0541 BPB - 5-expert Hedge Mixer + CROWN-Q + stride=64

RoyiRa:
# Record: 5-expert Hedge Mixer + CROWN-Q + stride=64 (val_bpb=1.0541)

**val_bpb: 1.0541** (3-seed mean) | **~15.7 MB** | 8xH100 SXM

## Results (8xH100 80GB SXM)

| Seed | step_avg | steps | Pre-TTT bpb | **Post-TTT bpb** | TTT gain | Eval time | Artifact |
|------|----------|-------|-------------|-----------------|----------|-----------|----------|
| 1337 | 98.1ms | 5,935 | 1.1251 | **1.0473** | -0.0778 | 336s | 15.89 MB |
| 42 | 97.9ms | 5,947 | 1.1264 | **1.0686** | -0.0578 | 336s | 15.69 MB |
| 7 | 98.0ms | 5,940 | 1.1246 | **1.0465** | -0.0781 | 336s | 15.66 MB |
| **Mean** | | | 1.1254 | **1.0541** | -0.0713 | 336s | ~15.75 MB |

### 1. CROWN-Q Training Penalty (training-time)
Added a quantization-aware penalty during warmdown that penalizes weights sensitive to quantization error:
```
crown_q_loss = lambda * mean(w^2 * delta^2 / 12)
```
where `delta = row_max / clip_range` is the per-row quantization step size. This encourages weights to be quantization-friendly, reducing post-quantization degradation. `CROWN_Q_LAMBDA=0.01`.

**Effect**: Slightly better compression (artifact ~200KB smaller) and more robust quantization.

### 2. Eval stride 32 -> 64 (eval-time)
Changed sliding window stride from 32 to 64 during evaluation. Experiment showed identical BPB quality but 2x faster scoring. Frees ~100s of eval budget for more TTT epochs.

### 3. TTT Epochs 3 -> 4 (eval-time)
Increased test-time training from 3 to 4 epochs per chunk, using the time freed by stride=64. Each additional epoch adapts the model more to scored data. Tested 8 epochs but that overfits (1.0735 vs 1.0473 for 4 epochs).

### Combined Effect
- stride=64 saves ~100s of eval time
- 4th TTT epoch uses ~85s of the saved time
- Net eval time: ~336s (down from ~562s), well within 600s budget
- BPB improvement: 1.0745 -> 1.0541 (-0.0204)

## Architecture

| Component | Setting |
|-----------|---------|
[...]

PR #461, opened 2026-03-22
Non-record: 11L Depth Recurrence + High-Yield Legal TTT (1.14458 BPB)

Christopher-Lee-McClendon:
## Non-Record Submission: 11L Depth Recurrence + Legal Score-First TTT

**val_bpb = 1.14458** | Pre-TTT: 1.1611 | TTT gain: **−0.0165** | Artifact: 14.79 MB

> Non-record unlimited-compute submission (trained on 4×A100-40GB).

---

### Headline

This submission demonstrates that **competition-legal test-time training can deliver large gains** when properly tuned. The key finding is a TTT recipe — SGD with momentum, multiple epochs per chunk, freezing early blocks — that extracts **2.4× more improvement** (−0.0165 BPB) than single-epoch AdamW over the full network (−0.0068 in our prior PR #456).

Every validation token is **scored before any weight update** that could use it, enforced by `torch.inference_mode()` during scoring.

### Novel & Creative Contributions

1. **High-yield legal TTT via selective freezing + SGD momentum** — Most TTT approaches use AdamW for 1 epoch over all parameters. We use SGD+momentum(0.9) for 3 epochs per 32K chunk while freezing the first 2 blocks. This is simpler, uses less memory, and gets 2.4× better TTT gains.

2. **Depth recurrence** — 11 logical layers from 10 unique BlockCores (one core reused at two depths with independent normalization). Delivers 11-layer capacity at 10-layer parameter cost.

3. **Partial RoPE (16/64 dims)** — Only 16 of 64 head dimensions use rotary embeddings, with NTK-aware scaling. The remaining 48 act as position-agnostic content channels, improving length generalization during TTT.

4. **Value Embeddings on deep layers only** — 128-dim learned embeddings added to value projections on layers 9–10, giving deep layers direct token-identity access in the value stream.

5. **Layer-Norm depth scaling** — `1/√(layer+1)` scaling stabilizes training under depth recurrence.

### Architecture

| Component | Setting |
|---|---|
| Layers | 11 logical (10 shared BlockCores) |
| Dim/Heads | 512 / 8 (4 KV heads) |
| MLP | 3× (1536), ReLU²; SmearGate |
| BigramHash | 2048 |
| RoPE | 16/64 dims, NTK scaling |
| Value Embed | 128d on layers 9–10 |
| XSA | Last 4 layers |
| Quant | Int6 + zstd |

### TTT Protocol

```
for each 32K chunk:
    1. model.eval() + inference_mode → score chunk (NLL recorded)
[...]

PR #825, opened 2026-03-26
Record: Order-Adaptive BackoffMixer (mean val_bpb=0.5440)

hypery11:
## Results

| Seed | val_bpb | Eval time |
|------|---------|-----------|
| 42   | 0.5437  | ~391s     |
| 1337 | 0.5450  | ~391s     |
| 2024 | 0.5434  | ~391s     |
| **Mean** | **0.5440** |           |
| Std  | 0.0008  |           |

- Artifact: ~16.0 MB
- Train: 600s on 8xH100 SXM
- Eval: ~391s (well under 600s)

## Method

11-layer transformer (512d, 8/8 full MHA, XSA-all, LeakyReLU(0.5)^2, 3.5x MLP). Order-adaptive entropy-gated BackoffNgramMixer with per-order entropy thresholds. Score-first, backward-looking, deterministic.

## Acknowledgments

Huge thanks to the incredible community that made this possible:

- @abaybektursun (PR #549) — base architecture + Legal TTT + Parallel Muon
- @deanbrr (PR #659, #779) — invented the n-gram eval cache for this competition, BackoffNgramMixer
- @Asukabot0 (PR #715, #727) — XSA-all concept, entropy-adaptive alpha formula
- @gowtham0992 (PR #606) — int5 + Soft-Round QAT
- @signalrush (PR #414) — EMA training recipe
- @sofiabod (PR #518) — LeakyReLU activation
- @thwu1 (PR #180) — mixed quantization, BigramHash, SmearGate
- @RoyiRa (PR #700) — TTT framework
- @Christopher-Lee-McClendon (PR #461) — TTT recipe
- @raahilshah (PR #162) — int6 quantization

This competition has been an amazing collaborative experience. Every improvement here builds on ideas shared openly.

- [x] 8xH100 SXM, train <=600s
- [x] Eval <=600s (391s)
- [x] Artifact <=16MB
- [x] 3-seed validation (std 0.0008)

### DIFF
--- a/records/track_10min_16mb/2026-03-26_OrderAdaptive_BackoffMixer/README.md
+++ b/records/track_10min_16mb/2026-03-26_OrderAdaptive_BackoffMixer/README.md
@@ -1,0 +1,38 @@
+# Order-Adaptive BackoffMixer
+
+## Results
+
+| Seed | val_bpb | Eval time |
+|------|---------|-----------|
+| 42   | 0.5437  | ~391s     |
+| 1337 | 0.5450  | ~391s     |
+| 2024 | 0.5434  | ~391s     |
+| **Mean** | **0.5440** | |
+| Std  | 0.0008  | |
+
+Artifact: ~16.0 MB. Train: 600s on 8xH100 SXM. Eval: ~391s.
+
+## Architecture
+
+11-layer transformer with XSA-all, full MHA (8/8 heads), LeakyReLU(0.5)^2, 3.5x MLP expansion. int5 quantization + compression. EMA, Tight SWA, Soft-Round QAT.
+
+## Eval-Time Scoring
+
+Order-adaptive entropy-gated BackoffNgramMixer with multi-order backoff (2-7 gram). Per-order entropy thresholds for mixing weight selection. Score-first, backward-looking, deterministic.
+
+## Acknowledgments
+
+Standing on the shoulders of many contributors to this competition:
+
+- @abaybektursun — PR #549 (base architecture, Legal TTT framework, Parallel Muon)
+- @deanbrr — PR #659, #779 (original n-gram eval cache concept, BackoffNgramMixer, drift-free TTT)
+- @Asukabot0 — PR #715, #727 (XSA-all, backoff concept, entropy-adaptive alpha)
+- @gowtham0992 — PR #606 (int5 + Soft-Round QAT)
+- @signalrush — PR #414 (EMA training recipe)
+- @sofiabod — PR #518 (LeakyReLU activation)
+- @thwu1 — PR #180 (mixed quantization, BigramHash, SmearGate)
+- @RoyiRa — PR #700 (TTT framework extensions)
+- @Christopher-Lee-McClendon — PR #461 (TTT recipe)
+- @raahilshah — PR #162 (int6 quantization baseline)
+
+This competition has been an incredible collaborative learning experience. Every improvement builds on ideas shared openly by the community.
--- a/records/track_10min_16mb/2026-03-26_OrderAdaptive_BackoffMixer/submission.json
+++ b/records/track_10min_16mb/2026-03-26_OrderAdaptive_BackoffMixer/submission.json
@@ -1,0 +1,9 @@
+{
+  "name": "Order-Adaptive BackoffMixer",
+  "val_loss": 0.5440,
+  "bytes_total": 15995959,
+  "blurb": "11L XSA-all 8/8 MHA, order-adaptive entropy gating + BackoffNgramMixer. Mean 0.5440 (std 0.0008) across 3 seeds. Eval 391s.",
+  "author": "hypery11",
+  "github_id": "hypery11",
+  "date": "2026-03-26"
+}
--- a/records/track_10min_16mb/2026-03-26_OrderAdaptive_BackoffMixer/train_gpt.py
+++ b/records/track_10min_16mb/2026-03-26_OrderAdaptive_BackoffMixer/train_gpt.py
@@ -1,0 +1,1777 @@
+"""V27: CROWN-Q training + stride=64 + 4 TTT epochs."""
+from __future__ import annotations
+import copy
+import glob
+import io
+import math
+import os
+import random
+import subprocess
+import sys
+import time
+import uuid
+import zlib
+import lzma
+from pathlib import Path
+try:
+    import zstandard
+    _COMPRESSOR = "zstd"
+except ImportError:
+    _COMPRESSOR = "lzma"
+import numpy as np
+import sentencepiece as spm
+import torch
+import torch.distributed as dist
+import torch.nn.functional as F
+from torch import Tensor, nn
+from torch.nn.parallel import DistributedDataParallel as DDP
+try:
+    from flash_attn_interface import flash_attn_func as flash_attn_3_func
+    _HAS_FA3 = True
+except ImportError:
+    try:
+        from flash_attn import flash_attn_func as flash_attn_3_func
+        _HAS_FA3 = True
+    except ImportError:
+        _HAS_FA3 = False
+        flash_attn_3_func = None
+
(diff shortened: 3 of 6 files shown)


MatoTeziTanka:
Circling back on this one with an updated finding, since @valerio-oai ruled on the underlying mechanism after my first comment.

**Compliance flag — same disallowed pattern as PR #779.**

@valerio-oai disallowed PR #779 (deanbrr) on 2026-03-27 ([comment 4145781641](https://github.com/openai/parameter-golf/pull/779#issuecomment-4145781641)) specifically for *"hashed n-gram caches, which do not renormalize correctly / correctly reweight the LM's token distribution, look ahead to the target token to mix probabilities and therefore leak eval tokens."* The mechanism is spelled out in the [follow-up comment 4146407380](https://github.com/openai/parameter-golf/pull/779#issuecomment-4146407380): hashing the ground-truth token into the lookup key only reweights the correct token, and in the hash-collision limit drives P(correct) toward 1 regardless of the data, giving arbitrarily low BPB without real compression.

Looking at `records/track_10min_16mb/2026-03-26_OrderAdaptive_BackoffMixer/train_gpt.py`, the `BackoffNgramMixer` (L39–145) is a port of #779's mixer with an entropy-gating delta on top, and it uses the same target-in-key hashing pattern at:

- **L76 (`update`):** `full_key = ((ctx_hash ^ (tgt * self.primes[cw])) & mask).astype(np.int64)` — hashes target `tgt` into the bucket
- **L78:** `np.add.at(self.full_counts[oi], full_key, 1)` — increments the target-conditioned count
- **L119 (`mix_and_score`):** `full_key = ((ctx_hash ^ (y_np.astype(np.uint64) * self.primes[cw])) & mask).astype(np.int64)` — same hash with `y_np` as the target
- **L121:** `full_c = self.full_counts[oi_rev][full_key.reshape(-1)]` — looks up the target-conditioned count
- **L1091:** `mixer.update(val_tokens[chunk_start_tok:chunk_end_tok + 1])` — also still has the `+1` boundary leak I flagged on #779 (which was fixed there in commit c58742a after my review; this PR branched from pre-fix code).

Under @valerio-oai's #779 ruling, this is the same Rule 1 violation (Issue #1017 condition 1 — `p_t` may depend only on the artifact and `x_1...x_{t-1}`). The 0.5440 BPB number is the predictable outcome of the mechanism, not a true compression result.

**Recommendation to @cocohearts @valerio-oai @0hq @yuzhougu-oai @notapplica:** **CLOSE** under the same ruling as #779. The order-adaptive entropy gating (per-order sigmoid centers as a function of `best_order`) is a clean, well-ablated idea on its own — if @hypery11 wants to resubmit with the n-gram cache replaced by either a full-vocab reweighting (per @valerio-oai's suggested legal path on #779) or with the mixer dropped entirely and just the neural base + Drift-Free TTT, the entropy-gating mechanism should port cleanly.

@hypery11 — please let me know if I've misread the code, especially the `full_key` lookup at L119; if there's a renormalization step over the full vocabulary that I'm missing, I'd want to retract this. Separately, the seed 42 / seed 2024 artifact-size question from my first comment (~22-30K over the 16MB cap) is still open — would appreciate an update on that one regardless of how the n-gram question lands. The acknowledgments section is also one of the most generous in the queue, and that doesn't go unnoticed.

---
*Reviewed by [@MatoTeziTanka](https://github.com/MatoTeziTanka) — [The Agora](https://matotezitanka.github.io/parameter-golf). Static code review against `train_gpt.py` at SHA `79ae889a`. AI tooling: review drafted with Claude Code (Sonnet/Opus) using an internal review template; all citations, file paths, and compliance audits were verified against the PR's actual code.*

Bortlesboat:
Positive compliance note from [parameter-golf-checker](https://github.com/Bortlesboat/parameter-golf-checker) — running across open Record-claiming PRs to help with triage (#1603).

Manually traced the chunked eval flow here and wanted to leave a clean note so it doesn't get bucketed with the TTT/SLOT cluster even though the n-gram trigger can look similar at a glance.

**What I verified (`eval_val_sliding_ttt` around lines 912–1127):**

The chunk loop at line 1020 does exactly what issue #1017 requires:

```python
# line 1020
for ci in range(num_chunks):
    ...
    # line 1025 — Phase 1: SCORE this chunk (inference_mode, no grad)
    base_model.eval()
    with torch.inference_mode():
        ...
        if mixer is not None:
            nll, expert_nll = mixer.mix_and_score(logits_scaled, x_batch, y_batch, wlens)
        ...
        # scoring accumulates into loss_sum / token_count / byte_count

    # line 1087 — Update context mixer with scored chunk tokens
    if mixer is not None:
        mixer.update(val_tokens[chunk_start_tok:chunk_end_tok + 1])

    # line 1098 — Phase 2: TRAIN on this chunk (already scored = legal)
    if not is_last_chunk and ttt_epochs > 0:
        ...
        optimizer.step()
```

For any token in chunk `k`:

1. `base_model` state is from training on chunks `[0, k-1]` only (the training step for chunk `k` runs at line 1098, *after* the score at line 1025).
2. `mixer` n-gram counts are from chunks `[0, k-1]` only (the `mixer.update` at line 1091 runs *after* the score).
3. Within chunk `k`, the per-position n-gram context at line 114–119 (`x_np[:, :slen - shift]`) uses strictly-prior tokens inside `x_batch`, which is the standard causal context the neural model also sees — no future-token leak.
4. Scoring happens under `torch.inference_mode()` at line 1038, so no gradients flow back into any parameter used to score these tokens.

The in-code comment at line 1098 (`# --- Phase 2: TRAIN on this chunk (already scored = legal) ---`) matches what the control flow actually does. The Polyak swap at 1030–1035 / 1093–1096 is for scoring stability only and does not change which chunks' data touched which state.

The 0.5440 BPB is striking but I don't think it implies a violation — a 7-gram backoff cache that grows over ~47M in-distribution val tokens is a legitimately strong mixer, and "score chunk k → update mixer with chunk k → score chunk k+1" respects the causality constraint at the chunk granularity.

**N-gram flag in my tool is a WARN, not a FAIL** — just want to flag that clearly in case the C3/N-gram warnings get batched with the actually-illegal cluster.

No action needed on this PR from me. Nice submission.

*(I've been wrong before — if I'm misreading something please push back.)*

deanbrr:
Hey, I started the entire n-gram hash thread and debate and while we can debate what 'learning' means in this contest, all the approaches have been ruled illegal. "The n-gram cache builds state from evaluation tokens and uses it to predict subsequent tokens. That's eval-time adaptation regardless of whether it's causal."

I do believe this is a very important topic because it presumes the approach is completely wrong when in fact the transformer itself is flawed in this regard:

The standard argument says: "the model should be fixed at eval time, any adaptation is cheating." That draws an arbitrary line.

Consider what a transformer does within its context window: it attends to prior tokens, builds key value representations, and uses them to predict the next token. That is learning from the eval data. It's building an internal model of the local distribution in real time. Nobody calls that illegal. We call it "in context learning" and use it.

The n-gram cache does the same thing with a longer window. A transformer with a 47M token context window would achieve similar benefits and nobody would call that illegal. The cache is just a more parameter efficient implementation of long-range context conditioning.

So the real question isn't "is the predictor fixed".  No useful predictor is fixed. Every autoregressive model conditions on previously seen tokens. The question is: where do you draw the line on context length and mechanism?

The competition draws it at "the 16MB artifact should be the complete predictor." But a transformer artifact without any context also predicts horribly. Every predictor requires input data to function.

I think the philosophical distinction is blurry, but there is a practical one. The competition measures "how much can you compress into 16MB of weights." The n-gram cache shifts the answer from "a good model of English" to "a good framework for memorizing any specific text." Those are different capabilities with a different value. Your point about "best predictor vs best specialist" captures this even if the mechanism is philosophically continuous with in context learning, the optimization target diverges.

It's a legitimate debate though, not a clear cut violation as the organizers have suggested.

PR #875, opened 2026-03-26
New Record: Pure Neural GDN 1.0226 BPB (shalyhinpavel)

shalyhinpavel:
### Summary
New SOTA record using **Gated DeltaNet (GDN)** architecture, achieving **1.0226 BPB** (average over 3 runs). This is a pure neural approach without any test-time training (TTT) or external caching.
 ### Key Changes
- **Architecture:** Replaced standard Attention with `GatedDeltaBlock` (Gated DeltaNet) layers.
- **Optimization:** Implemented a dynamic batch size and chunk size curriculum based on elapsed time.
- **Data Loading:** Switched to `FastLoader` with non-blocking prefetching and pinning.
- **Efficiency:** Achieved significant BPB improvement (~0.0968 nats) over the previous leaderboard record.

### Reproducibility
- **Artifact Size:** Under 16MB (zlib + int8 quantization).
- **Training Time:** ~10 minutes on 8xH100.
- **Logs:** 3 full runs provided in `records/track_10min_16mb/2026-03-26_Pure_Neural_GDN_1.0226/train.log`.


MatoTeziTanka:
### Community Review — Gated DeltaNet (GDN), Pure Neural

**BPB:** 1.0226 (claimed, withdrawn by author) | **Seeds:** 3 (42 / 1337 / 2024) | **Artifact:** ~14.1 MB claimed | **Compliance:** CONCEDED BROKEN by author

**What this does:** 8 layers of `fla.layers.delta_net.DeltaNet` (Gated Linear Attention) + 1 standard causal self-attention block on top, n_embd=384, n_head=6, vocab=1024, SP1024 tokenizer. Pure neural — no TTT, no n-gram cache, no SLOT, no eval-time mixers. Trains with AdamW (lr 1.8e-3, betas 0.9/0.95), a three-stage "golden ladder" batch/chunk curriculum (B=64/T=64 → B=128/T=128 → B=192/T=256), int8 row-scale quantization + DEFLATE zip. Declares `fla-core==0.4.2`, `flash-linear-attention==0.4.2`, `transformers==4.44.2` in `records/track_10min_16mb/2026-03-26_Pure_Neural_GDN_1.0226/requirements.txt`.

**What I found in the code** (`records/track_10min_16mb/2026-03-26_Pure_Neural_GDN_1.0226/train_gpt.py`):

1. **BPB constant is hardcoded, line 228:** `val_bpb = val_loss / (math.log(2) * 3.5)`. The submission scores `val_loss / (ln 2 · 3.5)` regardless of the tokenizer's actual token→byte ratio. Per README and Issue #897, BPB must come from real SentencePiece piece-length accounting (`base_bytes`, leading-space, boundary-token corrections); `3.5` is arbitrary and has no relation to the `fineweb_1024_bpe` vocab.

2. **`FastLoader` trains on a single training shard, line 177:** `self.data = np.memmap(files[0], dtype=np.uint16, mode='r')`. Only the lexicographically first `fineweb_train_*.bin` file is ever opened. All 2,320 training steps over 3 seeds drew from the same ~100M-token slice instead of the full 10B FineWeb split.

3. **Eval loads a single val shard and runs for a fixed 50 batches × 16 × 1024 ≈ 819,200 tokens, lines 198 & 224:** `val_path = os.path.join(base_data_dir, "fineweb_val_000000.bin")` and `for i in range(50): ...`. Evaluation never touches `fineweb_val_000001.bin` or beyond, and never runs a full sliding-window stride pass over the held-out set. @Eppie flagged the 50-batch cap upstream; it's still in the committed artifact.

4. **Cross-chunk DeltaNet state is dropped silently, lines 68–72:** `GatedDeltaBlock.forward` calls `self.delta_net(self.ln_1(x), state=state)`. `fla.layers.delta_net.DeltaNet.forward` (fla-core 0.4.x) takes `hidden_states`, `attention_mask`, `past_key_values`, `use_cache`, `output_attentions` — there is no `state` kwarg. `state` lands in `**kwargs` and is ignored. The outer loop in `main()` (lines 256–259) then calls `.detach()` on whatever is returned from `new_states`, but nothing recurrent is ever threaded in. With `chunk_size ∈ {64, 128, 256}` and `seq_len = 1024`, the recurrent state never crosses chunk boundaries — the defining property of a linear-attention RNN is absent from every forward pass of training and eval.

5. **Targets and inputs are clipped to the first 1024 ids, lines 184 and 221:** `np.clip(buf, 0, 1023, out=buf)`. The fineweb_1024_bpe tokenizer does emit ids in [0, 1023], so this is a no-op on the intended data, but it would silently corrupt any shard using a larger vocab — worth flagging as a latent footgun rather than a current bug.

6. **Roundtrip load uses `strict=False`, line 218:** `m.load_state_dict(new_sd, strict=False)`. Any key mismatch between the quantized state dict and the model will be swallowed. Given the custom `wte`/`lm_head` tying logic and the int8 dequant path, a silent partial load can't be ruled out.

7. **The log file numbers are not the "1.0226" BPB.** `records/.../train.log` prints `Final BPB: 1.0288 / 1.0198 / 1.0194`, averaging 1.0227, but each of those is the `val_loss / (ln 2 · 3.5)` output from the broken `official_judge` path above. They are not verified by any compliant evaluator.

**Smoke test (CT2038 proteus-engine, CPU, 2026-04-11):** `IMPORT_FAIL error=ModuleNotFoundError("No module named 'fla'")`. The script has a hard import of `fla.layers.delta_net.DeltaNet` at module top (line 7), so any eval environment without `flash-linear-attention` / `fla-core` 0.4.2 installed cannot even parse this submission. This is consistent with the PR's `requirements.txt`; flagging it only so reviewers know the container image will need those wheels for any re-run.

**Questions / flags:**
- Per Issue #897 and the README's tokenizer-agnostic BPB definition, the hardcoded `3.5` bytes-per-token constant means the reported number is not comparable to any other leaderboard entry.
- Per the README's validation protocol, evaluation must consume the full `fineweb_val` split with proper token→byte accounting. The 50-batch / single-shard loop here fails both conditions.
- Per the architecture's own premise, cross-chunk recurrence via DeltaNet state is the whole point of using a linear-attention RNN at this parameter count. The kwarg-name bug (`state=` vs `past_key_values=`) means the recurrence was never active during the reported runs.

**Author response:** In comment [#4149774501](https://github.com/openai/parameter-golf/pull/875#issuecomment-4149774501), @shalyhinpavel concedes all four bugs (single-shard train, single-shard eval, 3.5 BPB constant, dead DeltaNet state) and explicitly withdraws the leaderboard claim for this artifact: *"I concede the leaderboard claim for this specific artifact ... I will submit a mathematically and architecturally sound V2 shortly."* The author deserves credit for the fast, unambiguous concession.

**Verdict:** COMPLIANCE FLAG — reported BPB does not reflect the claimed architecture and does not use compliant byte accounting. Author has already conceded the record claim.

- CLOSE (or convert to non-record / draft) — the 1.0226 number cannot stand as a leaderboard entry given (a) hardcoded 3.5 BPB constant, (b) single-shard train and eval, (c) DeltaNet recurrence dead at the kwarg level, and (d) author's own concession in #4149774501. The Gated DeltaNet direction is genuinely interesting as a pure-neural baseline and worth encouraging in a V2 PR that rewires the recurrence (`past_key_values=` per fla-core 0.4.x), consumes all train/val shards, and uses the repo's standard `build_sentencepiece_luts` byte accounting.

---
*Reviewed by [@MatoTeziTanka](https://github.com/MatoTeziTanka) — [The Agora](https://matotezitanka.github.io/parameter-golf). CPU smoke test (CT2038 proteus-engine, 2026-04-11): IMPORT_FAIL — module `fla` (flash-linear-attention 0.4.2) not present in the smoke environment; this is a hard dependency declared in the PR's `requirements.txt`, not a script bug. AI tooling: review drafted with Claude Code (Sonnet/Opus) using an internal review template; all citations, file paths, and compliance audits were verified against the PR's actual code at SHA `90d5115e4071a2538e05dafeb6e97c9490d0afe8`.*

PR #708, opened 2026-03-25
[non-record] byte jepa-lw naive experimental baseline 2.125 bpb @ 2000 cc @0hq

juvi21:
adds a byte-level jepa naive baseline no tokenizer based on [le-wm](https://le-wm.github.io/)

in its current state this serves as a starting point encouraging experimentation and exploration, as mentioned in this [x thread](https://x.com/gum1h0x/status/2036558291451412560). there may still be small mistakes but it should be solid enough to build on. i may have some time and capacity on the weekend to continue working on this but would love to see what the oss community can build on it in meatime :) gl to anyone going for the swag! [ref](https://x.com/willdepue/status/2036440026863771682)

cc @0hq

### overview

it's basically it's le-wm with some small adaptations specific for this chall:

- byte-level input/output, no tokenizer byte 260
- exact next-byte compression objective added on top of the jepa loss.
- byte transformer encoder in place of the paper's pixel encoder
- action-conditioned prediction replaced with strictly causal chunk-context conditioning cuz byte compression has no external action stream
- predicted chunk latents injected into the byte scorer so jepa affects exact likelihood
- same core leworldmodel ingredients kept in the main path so latent prediction, batchnorm projector mlps, adaln predictor, and sigreg...
- EMA path kept only as a control

### current non-record run on rtx 5090

2.12515832 bpb @ 2000 steps

### how to replicate the logs:

```bash
uv venv
source .venv/bin/activate
uv pip install -r requirements.txt
uv run python data/cached_challenge_fineweb.py --variant byte260 --train-shards 1

RUN_ID=byte_jepa_896d14l_2000 \
DATA_PATH=./data/datasets/fineweb10B_byte260 \
MODEL_MODE=byte_jepa \
JEPA_ARCH=lewm_chunk \
JEPA_ENABLED=1 \
CHUNK_SIZE=32 \
TRAIN_BATCH_TOKENS=24576 \
VAL_BATCH_TOKENS=24576 \
TRAIN_SEQ_LEN=256 \
ITERATIONS=2000 \
VAL_MAX_BYTES=2000000 \
MODEL_DIM=896 \
NUM_LAYERS=14 \
NUM_HEADS=14 \
NUM_KV_HEADS=14 \
MLP_MULT=4 \
JEPA_LATENT_DIM=256 \
JEPA_PRED_HIDDEN=384 \
JEPA_PRED_DEPTH=2 \
JEPA_PRED_HEADS=4 \
JEPA_PROJECTOR_HIDDEN=2048 \
[...]

PR #906, opened 2026-03-26
[Research Non-Record] Pure raw-byte JEPA negative result

andrew-medrano:
## Summary

This PR documents the cleanest pure raw-byte JEPA attempt I ran for Parameter Golf. The best result was **`2.3839 bpb`** with `transformer_rope_gqa_localglobal + slot_ema_teacher`, which is a real improvement over my earlier pure-JEPA runs but still far from the simple baseline `1.2244`.

## What Makes This Pure JEPA

- raw `byte260`
- no tokenizer
- no exact byte-NLL into the backbone
- backbone trained only with JEPA-style latent prediction plus regularization
- exact byte prediction only through a later detached Transformer probe on frozen features

So the clean question here is narrow: can a pure raw-byte JEPA backbone, trained without exact-loss gradients, carry enough information that a later detached exact decoder can recover good `bpb`?

## Main Result

| Result | `bpb` | Notes |
|------|------:|------|
| Best pure detached-probe result | `2.3839` | `transformer_rope_gqa_localglobal + slot_ema_teacher` |
| Earlier purity-first milestone | `2.8583` | earlier raw-byte JEPA with a coupled exact decoder term |
| First clean frozen-probe milestone | `3.0774` | earlier pure-probe campaign |

## Controlled Comparisons

These were three fixed-budget comparisons:

1. **Backbone comparison**: same objective, same patch latent design, different Transformer backbones.
2. **Objective comparison**: same winning backbone, same patch latent design, different JEPA objectives.
3. **Patch-encoder comparison**: same winning backbone and objective, different within-patch latent encoders.

Headline winners:

| Comparison | Winner | `bpb` |
|------|------|------:|
| Backbone comparison | `transformer_rope_gqa_localglobal` | `2.3889800525604903` |
| Objective comparison | `slot_ema_teacher` | `2.3839` |
| Patch-encoder comparison | `conv_patch` | `2.746384624395377` |

## Comparison to Other JEPA PRs

| PR | Training path | Reported result | Why it differs |
|------|------|------:|------|
| This PR | pure raw-byte detached-probe JEPA | `2.3839` | no exact-loss gradients into backbone |
| [#708](https://github.com/openai/parameter-golf/pull/708) | hybrid JEPA + exact next-byte scorer | about `2.1252` | exact next-byte compression objective is in the main training path |
| [#896](https://github.com/openai/parameter-golf/pull/896) | tokenized JEPA self-distillation on top of autoregressive LM | PR author reports vanilla CE wins by `0.005 BPB` and is `40%` faster | useful negative result, but not raw-byte pure JEPA |
| [#903](https://github.com/openai/parameter-golf/pull/903) | LeWorldModel-style JEPA + SIGReg + CE head, plus a detached diagnostic probe | reported `1.2064` sliding / `1.2235` standard for best long BPE; `1.3348` standard for 10-minute byte | closest comparison, but the main reported model is still CE-trained and the JEPA-only contribution remains open |

PRs #708 and #896 are hybrid or auxiliary-loss approaches. PR #903 is closer to this line of work because it also includes a detached diagnostic probe, but its main reported model is still a CE-trained JEPA-augmented system rather than a pure backbone-only JEPA path. So none of them are apples-to-apples comparisons with this PR.

## Main Takeaways

- Stronger Transformer + slot targets improved pure JEPA a lot, but pure JEPA still remained far above baseline.
- Objective changes were small once the slot-target family was in place.
- Richer patch encoders mostly did not help.
- Lower JEPA loss did not reliably translate into lower exact-byte `bpb`.
- The main bottleneck now looks like latent/interface design, not just encoder size or JEPA loss choice.

## Why This Still Matters

This PR isolates the “pure JEPA” question more cleanly than the hybrid JEPA-related PRs in the repo. That makes it a useful lower bound and negative control for future JEPA claims: the best-performing JEPA-adjacent results still rely on a strong main CE path, which strengthens rather than weakens the negative result from the pure setup.


CiprianFlorin-Ifrim:
Good points, however, a few comments based on my work and things I found out through ablations:
1. My approach, just like the original LeWorldModel, exploits the temporal smoothness structure, not the cross view like the original LLM-JEPA paper. It's temporal latent prediction from LeWorldModel, predicting the next hidden state from the current one. LeWorldModel operates on single-stream video with no second view either, and it works because it enforces temporal smoothness and predictability in latent space via MSE + SIGReg. Now, the "translation" of this approach to text has to be further explored, as simpler method work better at this budgets (16MB), things will be different when the models are billions of parameters. 

2. There is a slight misunderstanding around the "code diff" example from before (which was just to give a reference to the fact that hybrid have been tested and work better, not that mine is based on that paper), I was referring to the underlying text itself. For example "text and sentiment" are 2 different views, same with someone saying "I want to create an SQL code to access my table X", it is the cross-view between the natural language and the SQL code. JEPA is able to capture the cross-view across all these tasks better than CE, which would miss a lot of nuances.

3. Through the ablations, it was found that a complex transformer or any other setup is not needed (at least at this size, can completly change with larger models), and add complexity without much benefit. In my code there is a probe that performs diagnostics to understand the behaviour of all the different elements (due to its hybrid nature).

On decoder complexity, through the ablations done it was found that a more complex decoder on worse representations loses to a simple decoder with better representations, and that there is a cap to the complexity needed for the decoder, and your work showcases that exactly. 

Diagnostics showed:
* Byte: lm_top1=0.70, lm_top5=0.91, lm_top10=0.96 with a simple tied linear head
* BPE: lm_top1=0.37, lm_top5=0.58, lm_top10=0.66 with a simple tied linear head

Top-k accuracy measures how often the correct next token appears in the model's k highest-ranked predictions, top1 means the model's single best guess is correct 70% of the time, top5 means the correct token is in the top 5 guesses 91% of the time. The final model uses a simple tied linear projection (no hidden layers, no nonlinearity) and reaches 1.2 BPB. The information gap is in the backbone, not the decoder.

MVPandey:
Ahhh gotcha, that's a fair distinction. Enforcing smooth latent trajectories is different from the multi-view setup I thoiught you were referencing. Agreed about size. CE might already be extracting everything the representation can hold at this scale.

MatoTeziTanka:
> **[RETRACTED 2026-04-11]** — This IMPORT_FAIL was a false positive. Root cause: Py3.10 @dataclass + spec_from_file_location harness bug. Your code is not broken. See correction below: https://github.com/openai/parameter-golf/pull/906#issuecomment-4230240338

---

### Community Review — [Research Non-Record] Pure raw-byte JEPA negative result

```
AttributeError: 'NoneType' object has no attribute '__dict__'
```

---
*Reviewed by [@MatoTeziTanka](https://github.com/MatoTeziTanka) — [The Agora](https://matotezitanka.github.io/parameter-golf). CPU smoke test (CT2038 proteus-engine, 2026-04-11): **IMPORT_FAIL** — AttributeError: 'NoneType' object has no attribute '__dict__'. Classification via `classify_prs.py` AST-based classifier; full compliance audit deferred until the import issue is resolved. Auto-drafted from a template and spot-checked before posting.*

MatoTeziTanka:
### Retraction — this IMPORT_FAIL was a Python 3.10 `@dataclass` loader bug in my harness

Sorry @andrew-medrano, this one's on me. I re-audited the `AttributeError: 'NoneType' object has no attribute '__dict__'` I reported above and confirmed it's a harness bug, not a bug in your code.

**Root cause:**

My smoke harness loaded your file with:

```python
spec = importlib.util.spec_from_file_location("train_module", script_path)
mod  = importlib.util.module_from_spec(spec)
spec.loader.exec_module(mod)   # ← crashed here
```

Python 3.10's `dataclasses.py` line 711 does `sys.modules.get(cls.__module__).__dict__` while processing your `@dataclass class Hyperparameters`. When a module is loaded via `spec_from_file_location` without first registering it in `sys.modules`, `cls.__module__` resolves to `"train_module"` but `sys.modules["train_module"]` is still `None`, so `.dict__` crashes. This is a well-known interaction between `importlib.util.spec_from_file_location` and `@dataclass` on 3.10 — fixed in 3.11+, worked around on 3.10 by registering the module name in `sys.modules` before `exec_module`.

**Verified at head `269432d`:**

Running your `records/track_non_record_16mb/2026-03-26_BytePatchJEPA_TransformerOnly/train_gpt.py` with the fix:

```python
sys.modules["train_module"] = mod
spec.loader.exec_module(mod)
```

…produces `IMPORT_OK, HAS_HYPERPARAMETERS=True, HAS_GPT=True`. Your code imports cleanly on Python 3.10.

**Your PR is not broken by this error.** I'm retracting the IMPORT_FAIL classification. I'll re-queue the full compliance audit (BPB check, n-gram / TTT / SLOT flags, etc.) and post findings separately.

Again — sorry for the noise. Harness bug, not your code.

PR #937, opened 2026-03-27
[Non-Record Submission] CompressedUT CE + EMA Export + Export-Aligned Late QAT (1.4457 BPB)

mihir-s-05:
## Summary
  This PR adds a non-record 16MB submission under `records/track_non_record_16mb/2026-03-27_CompressedUT_CE_EMA_ExportAlignedQAT_1.4457`. The submission explores a byte-level `compressed_ut` model trained in pure CE mode on `8xH100 SXM`, with the main goal of improving the *deployed* artifact rather than only the float-model validation number. The motivation for this line was that earlier versions of the model had a meaningful export gap: the full-precision model looked better than the quantized artifact that actually counts for submission. This run focuses on shrinking that gap.
  ## Architecture Motivation
  The core idea of the `compressed_ut` family is to avoid paying for full-resolution token processing at every layer.
  Instead of running a standard decoder stack over the full byte sequence, the model:
  - embeds raw bytes directly
  - builds local byte features with a lightweight convolutional frontend
  - compresses the sequence by stride into a smaller latent stream
  - runs the main transformer-style compute over that compressed latent stream
  - decodes bytes from the compressed latents with a small autoregressive byte decoder
  The motivation is to spend parameters and compute on a *smarter compressed latent backbone* rather than on a large full-resolution decoder. For a 16MB artifact budget, that tradeoff is attractive because it can preserve long-context reasoning while reducing the amount of per-token machinery that needs to be stored and quantized.
  This particular variant uses:
  - byte-level vocabulary (`VOCAB_SIZE=256`)
  - compressed latent backbone with `MODEL_DIM=1024`
  - widened byte embeddings (`BYTE_DIM=384`)
  - hashed bigram features (`BIGRAM_DIM=96`)
  - a small `slot_ar` byte decoder (`DECODER_DIM=192`, `DECODER_LAYERS=1`)
  - cross-conditioned latent-to-byte decoding inherited from the current `compressed_ut` path
  ## Why CE-Only
  Although `jepa/train_gpt.py` contains JEPA codepaths, this run intentionally uses:
  - `MODEL_FAMILY=compressed_ut`
  - `TRAIN_STAGE=ce`
  The motivation here was empirical: on this architecture line, CE-only was outperforming the joint JEPA variants we tested. So instead of continuing to push auxiliary JEPA losses, this submission doubles down on the strongest version of the deployed CE path.
  ## Main Ideas in This Submission
  The main design goal was **artifact-aware training**.
  Rather than just improving pre-quant validation, this run tries to make the exported int6 artifact itself better:
  1. **EMA export weights**
     - The exported model uses EMA-smoothed weights rather than the raw last-step weights.
     - Motivation: make the shipped model more stable and reduce last-step noise in the artifact.
  2. **Export-aligned late QAT**
     - Late in training, the model is exposed to the same quantization behavior used at export.
     - Motivation: train the model to be robust to the quantizer that actually defines submission performance, instead of letting quantization act as a post-hoc distortion.
  3. **Stronger int6 clip search**
     - Export tries a larger clip-search grid when packing weights.
     - Motivation: the compressed artifact budget still had headroom, so improving quantization quality was a better lever than only shrinking size further.
  4. **Larger compressed-UT capacity**
     - This run spends some of the remaining artifact budget on a stronger latent backbone and byte pathway.
     - Motivation: once the quantization penalty became manageable, extra capacity became a more attractive trade.
  ## Result
  Key numbers from the run:
  - pre-quant export-view eval: `val_bpb: 1.4386`
  - shipped artifact score: `final_quant_roundtrip_exact val_bpb: 1.44568091`
  - total submission size: `14,707,311` bytes
  - eval scope: full validation set
  - wallclock stop: `600.191s`
  - steps completed: `8694`
  A useful takeaway from this run is that the quantization penalty is relatively small:
  - `1.4386 -> 1.44568091`
  - roughly `+0.0071 bpb`
  That is the main success of this line: the quantized artifact now tracks the float model much more closely than earlier versions.
  ## Included Files
  This record includes:
  - `README.md`
  - `submission.json`
  - `train.log`
  - `train_gpt.py`
  - `launch_train_no_cudnn_sdp.py`
  - `requirements.txt`
  ## Notes
  - This is a non-record submission, not a leaderboard claim.
  - The included launcher wrapper disables cuDNN SDPA on H100 for this path, since direct launch had previously hit backend instability.
  - The logged artifact line was `zlib-9`, so the included README reproduces that exact path explicitly.


MatoTeziTanka:
### Community Review — [Non-Record Submission] CompressedUT CE + EMA Export + Export-Aligned Late QAT (1.4457 BPB)

Byte-level JEPA + EMA teacher, BigramHash legal. TTT disabled by default, legal when enabled.

---

PR #1145, opened 2026-03-30
Record: 1.1109 BPB FullGPTQ XSA11 + online (legal) ngram augment

AnirudhRahul:
## Summary
- build on [PR #1060](https://github.com/openai/parameter-golf/pull/1060)'s GPTQ XSA11 bigram-hash architecture
- retune the schedule with `WARMDOWN_ITERS=4000` and add a single-pass online token/within-word/word-start agreement evaluator packaged inside the record folder to improve bpb at eval time

## What `best_agree` Does
`best_agree` is a causal eval-time ensemble layered on top of the base model distribution.

It maintains three prefix-only experts:
- token n-gram top-token hints
- within-word continuation hints
- word-start first-token hints

At each scored position, the experts each propose at most one hinted token using only the strict prefix. The system then picks the best hinted token and applies a boost to that token inside the model's normalized distribution. When multiple experts agree on the same token, it adds a small extra agreement boost. So the gain comes from agreement between causal experts, not from looking up the gold token or rescoring with future information.

## Results
`val_bpb: 1.11085863` (4-seed mean, std `0.00030217`) | `15,953,221` bytes worst case | 8xH100 SXM

This improves the current README leader `1.1194` by `0.00592043 nats/byte` and `0.00854137 bpb` across four seeded runs.

A one-sided t-test confirms the improvement exceeds `0.005 nats/byte` over `1.1194` with `p = 0.00155` (`t = 8.7892`, `df = 3`), meaning there is only a 0.16% probability the observed gain is due to random chance.

| Seed | Standard sliding bpb | Online best-agree bpb |
|------|---------------------:|----------------------:|
| 42 | 1.11343872 | 1.11058356 |
| 1337 | 1.11408566 | 1.11126660 |
| 2025 | 1.11352210 | 1.11068499 |
| 15 | 1.11372333 | 1.11089935 |
| **Mean** | **1.11369245** | **1.11085863 (std 0.00030217)** |

## Why This Online Cache Is Valid
Earlier cache-style evals often failed because they either:
- queried the cache using the realized next token or let `x_t` influence whether a cache hit existed
- blended only the realized token probability instead of defining a full normalized vocabulary distribution
- treated bucket-local counts or hash-bucket scores as if they were already normalized token probabilities

This implementation is different:
- it uses only the strict prefix to choose at most one hinted token `h_t` plus a prefix-derived confidence before `x_t` is consulted
- it starts from the base model's full softmax distribution and applies a one-token boost with renormalization, i.e. `p'_t(a) = exp(beta_t * 1[a = h_t]) p_t(a) / Z_t`
- it scores position `t` before updating the online state with `x_t`
- it evaluates in a single left-to-right pass over the full validation stream, in order

So this is a causal, normalized online overlay on top of the base model rather than a target-conditioned or unnormalized cache score.

## Runtime
- 4-seed mean online eval wallclock: `467.78s` (std `9.06s`)
- the current implementation of the n-gram experts slows inference significantly, but this looks like an implementation issue rather than a fundamental limit and could likely be optimized a lot further

## Test plan
- [x] re-parsed the bundled seed logs and verified the README table fields against the logged metrics and total bytes
- [x] recomputed BPB-to-nats conversions, sample mean/stddev, 95% confidence interval, and the one-sided t-test directly from the four bundled logs
- [x] confirmed the bundled logs report online eval wallclock under the 10-minute budget on 8xH100

Across the different submissions and reruns I tried, these n-gram cache experts seem relatively consistent and typically give about a `0.003-0.004 bpb` boost.

### DIFF
--- a/records/track_10min_16mb/2026-03-29_Loader_FullGPTQ_XSA11_OnlineNgramAgreement/README.md
+++ b/records/track_10min_16mb/2026-03-29_Loader_FullGPTQ_XSA11_OnlineNgramAgreement/README.md
@@ -53,7 +53,7 @@ The implementation also keeps the metric calculation honest:
 The integrated online eval stays under the 10-minute evaluation budget on 8xH100.
 
 - 3-seed mean online eval wallclock: `468.34s` (std `11.01s`)
-- 8-GPU full-val benchmark log: `logs/bench_online_local_8gpu_full.txt`
+- 8-GPU full-val benchmark log: `online_eval_benchmark.log`
 - benchmark result: `1.11265002 -> 1.10955484 bpb` in `462.67s`
 
 The measured bottlenecks in the benchmark were the online overlay itself rather than the neural forward pass:
@@ -92,7 +92,7 @@ torchrun --standalone --nproc_per_node=8 train_gpt.py
 - `train_seed1337.log`
 - `train_seed2025.log`
 - `train_seed42.log`
-- `logs/bench_online_local_8gpu_full.txt`
+- `online_eval_benchmark.log`
 
 ## Credits
 
--- a/records/track_10min_16mb/2026-03-29_Loader_FullGPTQ_XSA11_OnlineNgramAgreement/README.md
+++ b/records/track_10min_16mb/2026-03-29_Loader_FullGPTQ_XSA11_OnlineNgramAgreement/README.md
@@ -1,19 +1,27 @@
 # Loader FullGPTQ XSA11 + Online Ngram Agreement
 
-**val_bpb: 1.11084505** (3-seed mean, std 0.00036858) | **15,995,106 bytes worst case** | 8xH100 SXM
+**val_bpb: 1.11085863** (4-seed mean, std 0.00030217) | **15,953,221 bytes worst case** | 8xH100 SXM
 
-Improves the current README leader at `1.1194` by **0.00592984 nats/byte** and **0.00855495 bpb** on the consistent 3-seed reruns.
+Improves the current README leader at `1.1194` by **0.00592043 nats/byte** and **0.00854137 bpb** on the bundled 4-seed subset (`42`, `1337`, `2025`, `15`).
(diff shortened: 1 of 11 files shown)


MatoTeziTanka:
### Community Review — Record: 1.1109 BPB FullGPTQ XSA11 + online (legal) ngram augment

**BPB:** 1.1109 | **Compliance:** LOOKS CLEAN — score-first-per-chunk TTT (legal #1416/#1423 pattern)

**What I found in the code** (head SHA `63cc3a350d1f`, file `records/track_10min_16mb/2026-03-29_Loader_FullGPTQ_XSA11_OnlineNgramAgreement/train_gpt.py`):

The TTT path at line **1214** implements the score-first-per-chunk pattern: each chunk is scored under `torch.no_grad()` / `inference_mode()` before the `base_model.train()` + SGD adaptation runs on that same chunk, with an `is_last_chunk` guard so the final chunk gets no adaptation pass. This is the structural shape the legal frontier uses (PRs #1416 erichroepke, #1423 aryanbhosale).

**CPU smoke test (CT2038 proteus-engine, 2026-04-11):** import OK in 0.16s, dim=512, layers=11, vocab=1024, code=103676 B, SMOKE_TEST_PASS

**Verdict:** LOOKS CLEAN.

---
*Reviewed by [@MatoTeziTanka](https://github.com/MatoTeziTanka) — [The Agora](https://matotezitanka.github.io/parameter-golf). CPU smoke test (CT2038 proteus-engine, 2026-04-11): import OK in 0.16s, dim=512, layers=11, vocab=1024, code=103676 B, SMOKE_TEST_PASS. Classification via deterministic AST-based `classify_prs.py` (pattern bank derived from ~65 manually-reviewed PRs earlier in the 2026-04-11 sweep). This review was auto-drafted from a template and spot-checked before posting — if the template misread your code, please call it out so I can iterate the classifier.*

PR #1125, opened 2026-03-30
Non-record: XSA-All + QK Gain 4.0 + LN Scale — 45 Experiments on 1×RTX 5090

jainpranjal97:
## Summary

Non-record submission: **1.1946 BPB** on 1×RTX 5090 (60-min, 3699 steps). 45 systematic experiments exploring hyperparameter space and novel architectures.

### Key findings for the community:

- **XSA on ALL layers > XSA on last 4** (-0.0018 BPB). Every top entry uses XSA only on deepest 3-4 layers — all-layer XSA is consistently better.
- **qk_gain_init = 4.0** (-0.0039 BPB vs default 1.5). Sharper attention patterns help small models significantly. Swept 1.5 → 2.0 → 3.0 → 4.0 with monotonic gains.
- **Warmdown calibration for wallclock-capped training** (-0.0078 BPB). Default warmdown_iters=1200 means the LR never reaches full strength when wallclock-capped at 10 min.
- **Pre-quant vs post-quant divergence**: XSA Gating (learned per-head gate) achieved 1.1932 pre-quant (better than best) but 1.1961 post int8+zlib (worse). Architectural choices that improve FP loss can degrade quantized loss.

### Novel approaches tested (all documented with negative results):

| Approach | Result | Why it failed |
|---|---|---|
| Progressive Layer Growing (5→11L at 60%) | +0.0057 | 5L capacity ceiling |
| Depth Recurrence 4×3 + LoRA16 | +0.0753 | torch.compile bypass + optimization conflicts |
| XSA Gating (learned per-head gate) | +0.0015 | Quantizes worse despite better FP loss |
| Cosine warmdown | +0.0039 | Linear warmdown already optimal |

### Stack

11L, MLP 3×, Partial RoPE 16/64, LN Scale 1/√(layer+1), XSA all layers, LeakyReLU(0.5)², Muon WD 0.06, seq 2048, grad_clip 0.3, qk_gain 4.0, logit_softcap 20.

Full experiment log with 45 runs in the README.

## Test plan

- [x] Verified val_bpb 1.1946 on 1×RTX 5090 (60-min run)
- [x] All 45 experiments logged with reproducible configurations
- [x] train_gpt.py included and runnable

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1089, opened 2026-03-29
Record Submission: 1.1091 BPB - Turbo-Muon + EngramLite + ParamBanking + XSA (11L 512d)

mikeapedia:
## Record Submission: 1.1091 BPB — Turbo-Muon + EngramLite + Parameter Banking + GPTQ Mixed-Precision

**val_bpb: 1.1091** (3-seed mean, std 0.0005) | **~15.3 MB** | 8×H100 SXM

### Results (8×H100 80GB SXM)

| Seed | step_avg | steps | val_bpb (SW s64) | val_bpb (roundtrip) | Artifact bytes |
|------|----------|-------|------------------|---------------------|----------------|
| 42   | 93.26ms  | 6284  | 1.1086           | 1.1324              | 15,992,528     |
| 1337 | 93.11ms  | 6295  | 1.1090           | 1.1328              | 15,993,413     |
| 2025 | 93.11ms  | 6294  | 1.1096           | 1.1335              | 15,993,904     |
| **Mean** | **93.16ms** | **6291** | **1.1091** | **1.1329** | |

### Summary

11-layer GPT (512d, 8H, 4KV GQA) combining eight key innovations over the PR #609 baseline:

- **Turbo-Muon Optimizer** — AOL preconditioning + Polar Express coefficients + row_col post-NS normalization (4 NS iterations instead of 5)
- **EngramLite** — Multi-head prime-based hash embeddings (bigram + trigram, 2 heads, 8192 buckets)
- **Parameter Banking** — 3D bank tensors enabling batched Newton-Schulz via `torch.bmm`
- **ASQU v3 Per-Layer Slopes** — Fixed per-layer LeakyReLU negative slopes from 3 rounds of adaptive tuning: layer 0 near-ReLU² (−0.014) → layer 10 (0.468)
- **U-Net Skip Connections** — Learned sigmoid-gated encoder/decoder skip paths
- **ValueEmbedding** — Token identity reinjection at deep layers (9, 10)
- **SmearGate** — Causal shift blending with predecessor token
- **XSA (all 11 layers)** — Efficient cross-sequence attention via GQA-aware reshape

### Architecture

Partial RoPE (16/64 dims), LN Scale (1/√(layer+1)), logit softcap 30.0, tied embeddings, per-head QK gain (init 1.5), LeakyReLU(ASQU v3 per-layer)² MLP at 3.5× width.

### Compression

[...]

PR #726, opened 2026-03-25
Memmap multi-shard data pipeline + GPU prefetch + LeakyReLU² + Legal TTT + Parallel Muon

DeepReinforce:
# Memmap multi-shard data pipeline + GPU prefetch + LeakyReLU² + Legal TTT + Parallel Muon

**val_bpb: 1.1147** (3-seed mean, std 0.0005) | **~15.23 MB** (mean int6+lzma + code) | 8×H100 80GB HBM3

## Results (8×H100 80GB HBM3, PyTorch 2.10.0+cu126)

| Seed | step_avg | steps | Pre-TTT bpb | **Post-TTT bpb** | TTT gain | TTT time | Artifact |
|------|----------|-------|-------------|------------------|----------|----------|----------|
| 1337 | 83.1ms | 7,223 | 1.1171 | **1.1140** | -0.0031 | 385s | 15,977,541 |
| 1111 | 83.0ms | 7,227 | 1.1178 | **1.1149** | -0.0029 | 383s | 15,964,369 |
| 13 | 83.0ms | 7,226 | 1.1179 | **1.1151** | -0.0028 | 386s | 15,957,041 |
| **Mean** | **83.1ms** | **7,225** | **1.1176** | **1.1147 (std 0.0005)** | **-0.0029** | **~384s** | |

*Pre-TTT bpb* is `final_int6_sliding_window_exact` (GPTQ-lite int6 dequantized weights, sliding window, stride 64). *Post-TTT bpb* is `legal_ttt_exact` after score-first chunk adaptation. *Artifact* is `Total submission size int6+lzma` (compressed weights + UTF-8 `train_gpt.py` bytes).

### 1. Training data pipeline

The old loader walked shards in fixed order and took contiguous spans per rank. The new `DistributedTokenLoader`:

- Memory-maps each `.bin` shard (`numpy.memmap`) with a small header-token cache to avoid repeated header reads.
- Samples **global** training windows across shards: per batch, it draws multiple shards with probability weighted by remaining usable blocks (exponent tightens as training progresses), uses a **coprime stride** over valid 2048-token blocks inside each shard, and interleaves windows for batch diversity.
- Merges nearby reads on the same shard into one slab copy to reduce mmap churn.
- Overlaps I/O and compute: a **daemon thread** builds CPU `(x, y)` batches into a queue while the main thread trains; **CUDA streams and events** prefetch the next batch to GPU while the current step runs.

[...]

PR #1218, opened 2026-04-01
Record: 4096-Vocab + 4.0-MLP-mult + 0.085-WD + Simplifications — val_bpb 1.09785 (3-seed mean)

clarkkev:
# Record: 4096-Vocab + Larger Model + High WD + Simplifications — val_bpb 1.09785

**val bpb: 1.09785** (3-seed mean, std=0.0004)

|-|-|-|-|-|-|
| 42   | 5967  | 1.10411 | 1.11588 | **1.09744** | 15,915,268 |
| 1337 | 5962  | 1.10482 | 1.11631 | **1.09795** | 15,905,460 |
| 2025 | 5961  | 1.10507 | 1.11641 | **1.09816** | 15,927,782 |
| **Mean** | | 1.10467 | 1.11620 | **1.09785** | 15,916,170 |

## Overview

This script builds on the 03-23 leaderboard [record](https://github.com/openai/parameter-golf/blob/main/records/track_10min_16mb/2026-03-23_LeakyReLU_LegalTTT_ParallelMuon/README.md). The main changes are:

### Fixes
* Fixed a small bug in the sliding window evaluation causing it to score tokens at the end of the val dataset multiple times. This bug didn't significantly affect results: it added roughly 2k duplicate contributions to the total loss and byte counts over a validation set of about 6M tokens. The faulty line was:
 `window_starts = [ws for ws in range(0, total_tokens, stride) if min(ws + seq_len, total_tokens) - ws >= 1]`, and it should be:
 `window_starts = [ws for ws in range(0, total_tokens, stride) if  ws + seq_len - stride  <  total_tokens]`

### Simplifications
* Use XSA in all layers instead of only the last 4.
* Removed parameter banking and distributed muon implementation and instead just used Muon + DDP.
* Removed test time training. I doubt that 0.1% additional tokens will improve the model generally, and for long docs I think it makes more sense to work on extending the sequence length.
* Removed quantization-aware training, since it appeared to provide little or no benefit.
* Removed gated attention.
* Removed value residuals.
* Removed hash embeddings, which are probably less necessary after increasing the vocab size.
* Removed the smear gate, for the same reason.

### Additions
* Increased the vocabulary size from 1024 to 4096. I used the existing `data/download_hf_docs_and_tokenize.py` to build the sentencepiece tokenizer and pre-tokenized data. The tokenizer model grew by ~50kb, but even with that added, the final artifacts would be below the 16MB cap. A larger vocab means the model sees more context for the same sequence length and more train data per step.
* Use a bigger but more strongly regularized model. I discovered that the compression ratio of a weight matrix (i.e., quantized-and-compressed-mb / raw-mb) correlates extremely well with the matrix's root-mean-square (`torch.sqrt(torch.mean(x**2))`) with an R^2 near 0.99. This suggests that the weight decay is a good lever for reducing the compressed size, which can let us add more parameters to the model. In particular this script uses:
  * Higher weight decays: muon weight decay increased 0.04 -> 0.085, and added an embeddings weight decay of 0.085. Additionally, decreased the adam weight decay 0.04 -> 0.02, as scalar parameters shouldn't need to be low-magnitude.
  * Wider MLPs, increasing `mlp_mult` 3 -> 4.
  * A decreased learning rate 0.025 -> 0.02, as larger models generally benefit from smaller LRs.
* Added the coprime-stride data loader from [#726](https://github.com/openai/parameter-golf/pull/726). The benefit is that it avoids showing the model sequences from the same document in the same/nearby minibatches by jumping around the data files.
* Added GPTQ Hessian-aware quantization. My implementation is based on [#1060](https://github.com/openai/parameter-golf/pull/1060) and reserves some time from training for Hessian computation.
* Use more efficient byte shuffle + brotli compression from [#1089](https://github.com/openai/parameter-golf/pull/1089).
* Added sigmoid-gated skip connections to the unet, also from [#1089](https://github.com/openai/parameter-golf/pull/1089).
* Increased `qk_gain_init` 1.5 -> 4 following [#1125](https://github.com/openai/parameter-golf/pull/1125).

### DIFF
--- a/records/track_10min_16mb/2026-04-01_Vocab4096_MLPMult4_WD085/README.md
+++ b/records/track_10min_16mb/2026-04-01_Vocab4096_MLPMult4_WD085/README.md
@@ -1,0 +1,87 @@
+# Record: 4096-Vocab + Larger Model + High WD + Simplifications — val_bpb 1.09785
+
+**val bpb: 1.09785** (3-seed mean, std=0.0004)
+
+| Seed | Steps | Pre-quant BPB | Post-quant BPB | **Sliding BPB** | Artifact |
+|-|-|-|-|-|-|
+| 42   | 5967  | 1.10411 | 1.11588 | **1.09744** | 15,915,268 |
+| 1337 | 5962  | 1.10482 | 1.11631 | **1.09795** | 15,905,460 |
+| 2025 | 5961  | 1.10507 | 1.11641 | **1.09816** | 15,927,782 |
+| **Mean** | | 1.10467 | 1.11620 | **1.09785** | 15,916,170 |
+
+## Overview
+
+This script builds on the 03-23 leaderboard [record](https://github.com/openai/parameter-golf/blob/main/records/track_10min_16mb/2026-03-23_LeakyReLU_LegalTTT_ParallelMuon/README.md). The main changes are:
+
+### Fixes
+* Fixed a small bug in the sliding window evaluation causing it to score tokens at the end of the val dataset multiple times. This bug didn't significantly affect results: it added roughly 2k duplicate contributions to the total loss and byte counts over a validation set of about 6M tokens. The faulty line was:
+ `window_starts = [ws for ws in range(0, total_tokens, stride) if min(ws + seq_len, total_tokens) - ws >= 1]`, and it should be:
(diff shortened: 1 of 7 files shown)


mikeapedia:
Awesome results @clarkkev!

PR #1224, opened 2026-04-01
Add non-record AR GPTQ XSA ROTQ Hadamard submission

vermissa0ss:
## Summary

  Adds a non-record submission folder for a rotation-aware GPTQ variant on top of the public AR self-generated GPTQ +
  XSA-all + BigramHash stack.

  Folder added:
  - `records/track_non_record_16mb/2026-04-01_ar_gptq_xsa_rotq_hadamard/`

  ## What this submission is

  This is a **non-record** submission.

  It uses the same evaluation metric and artifact accounting as the main leaderboard, but it is **not** a main-track
  record claim because the key run was trained on **1xH100 for 4800 seconds**, not reproduced under the official
  **8xH100 SXM / 600 second** budget.

  ## Main idea

  The new ingredient is modular rotation-aware GPTQ:
  - choose a compact right-rotation for selected large MLP matrices before GPTQ
  - quantize in the rotated basis
  - invert the rotation after dequantization during roundtrip evaluation

  The best result here used per-layer Hadamard right-rotations on `mlp_up` and `mlp_down` with block-size search over
  `{128, 256, 512}`.

  ## Results

  Best result in this folder:
  - exact sliding `val_bpb`: `1.11290586`
  - exact sliding `val_loss`: `1.87908996`
  - total artifact bytes: `15,826,148`
  - hardware / duration: `1xH100`, `4800s`, seed `314`

  Export-only ablations on the same checkpoint:
  - base export: `1.11296252`
  - `mlp_down` Hadamard `256`: `1.11291713`
  - `mlp_down` per-layer Hadamard: `1.11290938`
  - `mlp_up + mlp_down` per-layer Hadamard: `1.11290586`

  So the best rotation-aware export improves the same-checkpoint export path by `-0.00005666` BPB while staying under
  the 16MB cap.

  ## Notes

  - README is explicit that this is a non-record submission.
  - Included logs are recovered copies of the exact auto-generated pod log content because the pod stopped before raw
  log sync completed.
  - This PR only adds the new records folder.


cocohearts:
nice

PR #1106, opened 2026-03-30
Non-record: MDLM Diffusion — val_var_bpb 1.1465 (first diffusion to beat AR baseline)

agalimova:
## Summary

**val_var_bpb: 1.1465** (512 eval steps) | **33M params** | 2xH100 80GB HBM3 | Non-record

First discrete diffusion model to beat the AR baseline (1.22 BPB). Beats previous best diffusion (#820, 1.625; #1053, 1.360) by 0.21+ BPB.

## Results

| Model | BPB |
|-------|-----|
| AR SOTA (merged #1) | 1.1194 |
| **This (MDLM)** | **1.1465** |
| AR baseline | 1.2244 |
| #1053 MDLM | 1.360 |
| #820 MDLM | 1.625 |

## Approach

MDLM (Sahoo et al. 2024) with log-linear noise, adaLN timestep conditioning, frozen visible-token logits, antithetic sampling, discrete absorbing-mask ELBO eval. 11L 512d, 6000 steps, AdamW.

## Key findings from 27 experiments

- Masking eps=0.1 >> 0.001 (biggest win)
- Eval method matters: MC ELBO = 2.41 BPB, discrete ELBO = 1.15 (same model)
- AR tricks that don't transfer: LeakyReLU^2, BigramHash

## Hardware

Developed on GB10 (Project DIGITS). Validated on 2xH100 (TensorPool, 31 min). 8xH100 unavailable (#821). Extrapolated: ~8 min on 8xH100.

Issue #897, opened 2026-03-26
BUG: bpb underestimated when tokenizer does not contain U+2581 (ie the space) token


riccardoalberghi:
build_sentencepiece_luts (train_gpt.py:180) tries to estimate how many UTF-8 bytes each token corresponds to.

For normal tokens, it removes the leading ▁, treats that as a single space byte, and then counts the UTF-8 bytes in the rest of the token string. For byte-fallback tokens (sp.is_byte()), it just counts 1 byte per token.

The issue shows up when ▁ (U+2581) is not present as its own token. In that case, SentencePiece encodes it as three separate byte tokens: <0xE2>, <0x96>, and <0x81>, since ▁ is three bytes in UTF-8. Because each of those fallback tokens is counted as 1 byte, the code ends up treating ▁ as 3 bytes.

But in practice, ▁ is only used to represent a single ASCII space (0x20), which is just 1 byte. So every word boundary gets overcounted by 2 bytes.

That makes val_byte_count too large, which then makes tokens_per_byte look smaller than it really is and pushes BPB artificially low. I have run a few tests, we are talking roughly about a 20% decrease in reported bpb compared to the true one.

The standard fineweb_1024_bpe.model does include ▁ as token 939, so the normal stripping logic handles it correctly and this edge case never happens. It only appears with a custom SentencePiece model that does not contain a standalone ▁ token.

Hope this helps.
Riccardo

MatoTeziTanka:
The corrected tokenizer and pre-tokenized shards are now published.

**Scylla v2** — a byte-exact tokenizer that fixes the BPB underestimation described in this issue — is available at **[LightSpeedUp/parameter-golf-data](https://huggingface.co/datasets/LightSpeedUp/parameter-golf-data)** on HuggingFace:

```bash
huggingface-cli download LightSpeedUp/parameter-golf-data --include "fineweb_scylla_v2/*" --local-dir ./data
```

Includes the corrected `.yaml` tokenizer config, `.meta.npz`, and 81 pre-tokenized shards (~16.9 GB). The tokenizer path is the same one in PR #1314.

This bug was one of the reasons we built the full dataset drop — while fixing the Scylla tokenizer we realized none of the larger vocab variants (sp4096, sp8192, sp12288, sp16384, byte260) were available on HuggingFace either. All six variants are now in the same repo (~89 GB total).

---
[...]

PR #1241, opened 2026-04-02
MDLM Diffusion — val_var_bpb 0.9901, EOS learning + full dataset shard rotation, 33M params, 1x AWS A10G

aiejvn:
Builds on PR #1106 (MDLM stack). Two additions:

EOS learning: Token 1 (\<s\>) is used as a document boundary anchor — never masked during diffusion\. A dedicated PAD\_ID\=1025 \(separate from MASK\_ID\=1024\) fills post\-EOS positions and is excluded from the loss, preventing collision between structural padding and diffusion masking\.

Shard rotation: ShardedDataLoader loads N shards at a time and rotates between groups across training, enabling full FineWeb 10B training without loading the entire dataset into RAM. Explicit memory freeing between groups; shards loaded one-at-a-time into a pre-allocated buffer to avoid 2× peak allocation.

Ablation finding: Val BPB is flat across attention head counts {2, 4, 8, 16, 32} at fixed model dim — head count appears invariant for bidirectional diffusion LMs.

Non-record reason: Trained on 1× AWS A10G (1267 min). Requires 8×H100 SXM for wall-clock compliance.

| Model                             | BPB    |
|----------------------------------|--------|
| This (MDLM v5)                   | 0.9901 |
| PR #1106 (prior best diffusion)  | 1.1465 |
| AR baseline                      | 1.2244 |

### DIFF
--- a/records/track_non_record_16mb/2026-04-01_MDLM_EOS_FullDataset_A10G/README.md
+++ b/records/track_non_record_16mb/2026-04-01_MDLM_EOS_FullDataset_A10G/README.md
@@ -1,0 +1,133 @@
+# MDLM v5: EOS Learning + Full Dataset Shard Rotation
+
+**val_var_bpb: 0.9901** (128 eval steps) | **33M params** | 1x AWS A10G (24GB) | Non-record
+
+Builds on PR #1106. First MDLM submission to train on the full FineWeb SP-1024 dataset
+via shard rotation, and the first to learn document-boundary structure (EOS tokens) during diffusion.
+
+## Results
+
+| Model | BPB |
+|-------|-----|
+| AR SOTA (PR #549) | 1.1194 |
+| **This (MDLM v5, EOS + full dataset)** | **0.9901** |
+| PR #1106 (MDLM, prior best diffusion) | 1.1465 |
+| AR baseline | 1.2244 |
+
+## Why Non-Record
+
+Trained on 1x AWS A10G (24GB VRAM). The competition requires 8xH100 SXM within a
+10-minute wall-clock budget. The A10G was used for iterative development and to
+validate the full-dataset shard rotation strategy. We plan to rerun on 8xH100 when available.
+
+## What's New vs PR #1106
+
+PR #1106 established the MDLM baseline: bidirectional transformer, log-linear noise schedule,
+discrete absorbing-mask ELBO, frozen visible-token logits. This submission adds two
+contributions on top of that stack.
+
+### 1. EOS Token Learning
+
+In the SP-1024 tokenizer, token 1 (`<s>`) marks document boundaries throughout FineWeb.
+Prior MDLM submissions treated the corpus as a flat token stream with no document structure signal.
+
+**Key design:**
+- `EOS_ID = 1` — never masked during forward diffusion. EOS positions are always visible
+  to the model regardless of noise level, acting as structural anchors.
+- `PAD_ID = 1025` — a dedicated padding token (not MASK_ID=1024) fills positions after EOS
+  within each sequence. PAD is also never masked and is excluded from the loss.
+- Separating PAD from MASK avoids a collision where the model cannot distinguish
+  diffusion masking from structural padding.
+
+**Loss:**
+
+```python
+is_special = (x0 == EOS_ID) | (x0 == PAD_ID)
+move = (torch.rand_like(x0.float()) < move_chance[:, None]) & ~is_special
+xt = torch.where(move, MASK_ID, x0)
+
+# x0_safe avoids -inf * 0 = nan at PAD positions during gather
+x0_safe = x0.masked_fill(x0 == PAD_ID, 0)
+log_p_x0 = torch.gather(log_probs, -1, x0_safe[..., None]).squeeze(-1)
+content_mask = (x0 != PAD_ID).float()  # loss over real tokens + EOS only
+
+loss = (dsigma[:, None] * (-log_p_x0) * is_masked * content_mask).sum() / n_content
+```
+
+**Document chunking:** Documents are split into contiguous chunks of at most `SEQ_LEN=2048`.
+Short documents → one chunk ending at EOS + PAD fill. Long documents → N full SEQ_LEN chunks
+(no EOS mid-doc) + one tail chunk ending at EOS. EOS supervision is preserved at every
+document boundary; batch shapes remain consistent.
+
+### 2. Shard Rotation for Full-Dataset Training on Memory-Constrained Hardware
+
+The FineWeb SP-1024 training set spans 80 shards (~100M tokens each). Loading all 80
+simultaneously requires ~64GB RAM — infeasible on most single-GPU setups.
+
+**Strategy:** `ShardedDataLoader` partitions shards into groups of `SHARDS_IN_MEMORY`
+(default 4). Training runs `TRAIN_STEPS` steps per group before rotating, ensuring every
+shard is visited once per pass. Memory is explicitly freed between groups:
+
+```python
+# Explicit free before allocation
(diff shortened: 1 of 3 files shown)


MatoTeziTanka:
### Community Review — MDLM Diffusion v5 (EOS Learning + Shard Rotation)

**Reported:** `val_var_bpb` 0.9901 | **Seeds:** [42] (1 seed) | **Hardware:** 1x AWS A10G (1267 min) | **Track:** non-record

**What this does:** Builds on PR #1106's Masked Discrete Language Model (MDLM) baseline. Adds (a) document-boundary anchoring — token 1 (`<s>`) is treated as EOS and is *never* masked during forward diffusion, with a dedicated `PAD_ID=1025` (separate from `MASK_ID=1024`) filling post-EOS positions and excluded from the loss; and (b) `ShardedDataLoader`, which rotates groups of N FineWeb shards through RAM so that the full 80-shard SP-1024 corpus can be visited on a 24GB consumer/cloud GPU. Also reports a head-count ablation showing val BPB is roughly invariant to `n_heads ∈ {2,4,8,16,32}` for bidirectional diffusion LMs at fixed `MODEL_DIM=512`.

**Smoke test (CT2038 proteus-engine, 2026-04-11):** `IMPORT_OK seconds=0.02`. Top-level parses cleanly on CPU. The harness's `HAS_HYPERPARAMETERS` / `HAS_GPT` checks are not applicable here — this is a `DiffusionLM` module, not the AR `GPT`/`Hyperparameters` template. No top-level CUDA assumptions in module scope.

**Architecture (factual, from `train_mdlm.py`):**
- 11 layers, `MODEL_DIM=512`, 8 heads, MLP mult 3× (ReLU²), RoPE, AdaLN sigma conditioning, bidirectional `is_causal=False` SDPA. ~33M params (header).
- `TOTAL_VOCAB = 1026` (1024 SP + MASK + PAD); `PADDED_VOCAB = 1088` for the embedding/unembed table; `head(...)[..., :TOTAL_VOCAB]` slices outputs.
- `subs_log_probs` blocks `MASK_ID` and `PAD_ID` from prediction (`-1e6`), renormalises, and freezes visible positions to identity — standard MDLM substitution parameterisation.
- Training loss: continuous-time MDLM ELBO with antithetic time sampling, `dsigma * -log p(x0|xt)` masked to mask-only positions, weighted by `content_mask = (x0 != PAD)`.
- Eval: `variational_elbo_bits` runs a 128-step Riemann discretisation of the discrete absorbing-mask ELBO over 500 fixed `SEQ_LEN=2048` slices of the val split.

**Central question — is `val_var_bpb` directly comparable to the leaderboard metric?**

This is the core issue I'd like the author / mods to weigh in on. Two specific concerns visible in the code, both at the eval step:

1. **Variational upper bound, not exact NLL.** The reported number is the discrete absorbing-mask ELBO computed by `variational_elbo_bits` (`train_mdlm.py` ~L173–215). For diffusion LMs this is a *variational upper bound* on the true negative log-likelihood. The leaderboard metric, by contrast, is the exact per-token NLL of an autoregressive factorisation evaluated via sliding-window stride-64 on the canonical eval path. ELBO ≥ true NLL, so 0.9901 is an *upper* bound on this model's BPB under the MDLM factorisation — the actual NLL could be lower. But the inverse is also true: it is not the same quantity that AR submissions report. A 0.9901 ELBO from a diffusion model and a 0.9901 sliding-window NLL from an AR model are not interchangeable, and at minimum the comparison to "AR SOTA 1.1194" / "AR baseline 1.2244" in the README needs that asterisk attached. (See #1461 for an analogous metric-definition discussion on a non-AR submission.)

2. **Bytes/token uses a hardcoded `4.3` constant, not the per-token SP byte LUT.** Both the inline progress estimate and the final reported number compute `bpb = total_bits / (total_content * 4.3)` (`train_mdlm.py` L361, ~L398, ~L407). The competition scoring formula divides total bits by the *exact* sum of UTF-8 bytes per token from `build_sentencepiece_luts` over the eval slice — not by `n_tokens × avg_bytes_per_token`. For SP-1024 on FineWeb, average bytes/token is roughly in the 4.2–4.4 range, so 4.3 is in the right neighbourhood, but it is not what the canonical evaluator computes. The README claims "Competition BPB uses sentencepiece byte-count LUTs (exact bytes per token, matching competition scoring formula)" — but the code in this PR does not actually do that; it uses the constant. Whether the resulting number is high or low vs the canonical computation depends on the exact byte-distribution of the 500 eval slices, and is not something I can determine without running it.

In short: the headline 0.9901 is (a) a variational upper bound on the MDLM-factorised NLL, (b) computed against a hardcoded 4.3 bytes/token denominator rather than the exact SP byte LUTs, and (c) computed on 500 fixed contiguous `SEQ_LEN=2048` slices of val rather than the canonical sliding-window stride-64 path. Each of those is a separate axis of incomparability; the net effect is that this number cannot be placed on the leaderboard against AR submissions without a mod ruling on what "BPB" means for a diffusion LM, and ideally a re-run through the canonical evaluator.

**Other observations (not blockers, just notes):**
- This is correctly filed as **non-record**, and the PR body and submission.json both flag the A10G hardware and the need to rerun on 8xH100 SXM. So none of the above is a record-track compliance issue — it's a "what does the number on the title actually mean" question for the non-record / novel-architecture track.
- `submission.json` reports a single seed (`42`); the README's invariance ablation is across head counts at one seed, not across seeds for the headline number.
- No custom tokenizer; SP-1024 is the standard path, so #897 / U+2581 byte-fallback concerns do not apply.
- TTT / SLOT / n-gram cache rulings (#402, #677, #1336) do not apply — this is pure pretraining + ELBO eval, no test-time adaptation, no eval-time caches.
- The `MDLM_EOS_FullDataset_A10G/` directory adds a `train_mdlm.py` rather than the canonical `train_gpt.py` — worth noting that any automated runner wired to `train_gpt.py` will not pick this up as-is.
- Head-count invariance ablation is a genuinely interesting empirical claim for bidirectional diffusion LMs and worth keeping in the record regardless of the metric question.

**Verdict:** QUESTIONS RAISED — interesting architecture and a clean MDLM extension, but the headline `val_var_bpb` 0.9901 is a variational upper bound computed with a hardcoded bytes/token constant, not the canonical sliding-window stride-64 NLL the leaderboard uses. Worth keeping in the non-record track for the technique, but the number should not be compared directly to AR submissions without a mod ruling and ideally a canonical-evaluator re-run.

- HOLD on any leaderboard placement pending a ruling on (a) whether ELBO-based BPB is comparable to exact-NLL BPB for the non-record track and (b) whether the `4.3` bytes/token constant in the eval path needs to be replaced with the SP byte LUT before the number is quoted alongside AR submissions.
- NEEDS AUTHOR ACTION (optional, if author wants the number to be directly comparable): swap `total_content * 4.3` for `sum(byte_lut[t] for t in eval_tokens)` and re-run the 128-step ELBO; ideally also report the exact-NLL via the canonical sliding-window path on the same checkpoint, so both numbers are on the table.
- Otherwise the technique itself (EOS anchoring, PAD/MASK separation, shard rotation, head-count invariance ablation) is a clean and well-documented contribution to the diffusion-LM line and worth keeping in the non-record track on its own merits.

---
*Reviewed by [@MatoTeziTanka](https://github.com/MatoTeziTanka) — [The Agora](https://matotezitanka.github.io/parameter-golf). CPU smoke test (CT2038 proteus-engine, 2026-04-11): IMPORT_OK 0.02s; HAS_HYPERPARAMETERS/HAS_GPT N/A (DiffusionLM, not AR template). AI tooling: review drafted with Claude Code (Sonnet/Opus) using an internal review template; all citations, file paths, and compliance audits were verified against the PR's actual code at SHA `5d46f39f69290735adc989930c7f5fe4e0779158`.*

PR #1255, opened 2026-04-02
Non-record Submission: Text Diffusion + Retrodiction + TTT + Depth Recurrence

akaiHuang:
## Summary
This PR adds a non-record 16MB submission under records/track_non_record_16mb/2026-04-02_Meadow_TextDiffusion_Retrodiction_TTT_DepthRecurrence.

Techniques included:
- Text Diffusion (CDM) with Sequential Unmasking eval
- AR Retrodiction
- Test-Time Training (full-model AdamW)
- Depth recurrence experiments
- Custom v4096 tokenizer

## Files
- README.md
- submission.json
- train_gpt.py
- train_cdm.py
- eval_sequential_unmasking.py
- eval_ttt.py
- bpe_v4096.model
- train.log

This is intended for the non-record track.


MatoTeziTanka:
### Community Review — Non-record Submission: Text Diffusion + Retrodiction + TTT + Depth Recurrence

```
```

---

akaiHuang:
Thanks for the smoke-test details and for catching this before the official audit.

The four files that triggered the import error (`eval_cf_dualbrain.py`, `eval_ttt.py`, `eval_sequential_unmasking.py`, `leakage_test.py`) are Apple-Silicon-only pre-flight artifacts referenced in §3.3 / §3.4 / §3.6 of the README. They are not part of the H100 production training or evaluation path (which is `train_cdm.py` + `eval_cf_ablation.py`), but they live in the same PR folder so your folder-level smoke test correctly hit them.

**Fix pushed in commit `a4286a4`:** each file now wraps its `import mlx` in a `try / except ImportError` block, installs a minimal stub class on failure (so module-level definitions like `class Foo(nn.Module):` and `COMPUTE_DTYPE = mx.bfloat16` still parse), and sets `_HAS_MLX = False`. `leakage_test.py` additionally moves its module-level test logic into a `_main()` function gated by `if __name__ == \"__main__\":` so that importing the file is a no-op.

**Locally verified:**
- `py_compile` passes on all four files under Python 3.10 syntax.
- Import succeeds on a real Apple Silicon machine with MLX installed (`_HAS_MLX=True`, original behaviour preserved).
- Import succeeds on a simulated mlx-blocked environment via a custom \`__import__\` override (`_HAS_MLX=False`, stub path).

No README, no scoring path, no numbers change. Please re-run the audit at your convenience — happy to help debug if anything else trips.

MatoTeziTanka:
Re-audited at head SHA `a4286a4`.

**Fix confirmed.** The four MLX-dependent files (`eval_cf_dualbrain.py`, `eval_ttt.py`, `eval_sequential_unmasking.py`, `leakage_test.py`) all now wrap `import mlx` in `try/except ImportError` with `_HAS_MLX = False` fallback and minimal stub classes. `leakage_test.py` moves its module-level logic into a `_main()` gated by `if __name__ == "__main__"`. All files compile clean under Python 3.10.

As you noted, these are Apple Silicon pre-flight artifacts — not part of the H100 scored eval path (`train_cdm.py` + `eval_cf_ablation.py`). The fix is the right approach: guarded import so folder-level smoke tests don't false-positive while preserving full functionality on actual Apple Silicon machines.

The `train_gpt.py` (H100 path) compiles clean. I'll queue the full compliance audit on the active eval path for the next sweep.

---
*Re-audit by [@MatoTeziTanka](https://github.com/MatoTeziTanka). Verified MLX import guards in all 4 files, py_compile OK under Python 3.10.*

PR #1240, opened 2026-04-02
Non-record: Does SLOT violate causal dependence? (empirical test + question)

andrewbaggio1:
## What this is

I wrote a script with a coding agent that tests whether SLOT violates causal
dependence. I think the answer is yes.

## My Test

- Take a window of tokens, run SLOT, and record the NLL at each scored position.
- Flip a single target token somewhere in the window and re-run SLOT.
- If the NLL changes at other positions, those predictions depended on the token
  we flipped.

Without SLOT, changing a target has no effect on other positions because the
model is causal. With SLOT, every single scored position is affected, and I
found a 100% violation rate across 240 tested pairs.

I also checked self-prediction: is P(x_{t+1}) better when x_{t+1} is actually
in the optimization targets vs when it's swapped for a random token? Yes, by
+0.24 nats (shared delta) and +0.73 nats (per-sample + logit bias).

I'm flagging my own submission (PR #1209).

I looked at PR #1229, which pushes SLOT to its logical extreme: per-sample
delta, per-sample logit bias, and scored-position masking. It gets 0.9300 BPB,
which is the best on the non-verified leaderboard and 0.19 below merged SOTA.
In defense of it, the mechanism is the same as shared-delta SLOT, just with
more parameters to memorize the evaluation targets.

## Counterargument

@AnubhavBharadwaaj correctly points out that in stride=64 sliding window,
1984/2048 tokens are already-scored context. So ~97% of the gradient comes
from known tokens. I think this is fair because the leakage in shared-delta
SLOT is small. But it's not zero, and "a little bit of future information"
is still future information.

## Reproducing

```bash
# No GPU needed. ~30 seconds on CPU.
python prove_slot_causal_violation.py
```

## Request for a decision

@0hq @valerio-oai SLOT has been debated across PRs #1084, #1128, #1172,
#1176, #1209 without a ruling. I'd really appreciate you weighing in!

Full writeup in the README generated by Claude Code.

PR #1186, opened 2026-03-31
Non-record: Negative Results — Architecture, TTT Variants, Quantization, and N-gram Cache Illegality

andrewbaggio1:
## Summary

~15 experiments that didn't work or were marginal on the LeakyReLU(0.5)² stack (PR #518 architecture). Documenting these so others don't repeat them.

**Architecture:**
- Depth recurrence (Huginn-style): +0.20 BPB — compute overhead dominates
- TrigramHash: +0.045 BPB — quantization destroys small weights
- MLP 3.25x: marginal gain but artifact exceeds 16MB
- XSA-all (11 layers): neutral at MLP 3.0

**TTT variants:**
- SGD+momentum: +0.065 BPB worse than AdamW — adaptive LR matters
- MLP-only TTT: +0.237 BPB — needs meta-learning to work
- TTT LR floor 0.05: +0.015 BPB — cosine should decay to 0
- 20 vs 30 epochs: 10 extra epochs worth ~0.03 BPB

**Quantization:**
- int5 post-training swap: catastrophic — must train with int5 QAT
- 1xH100 training: not viable proxy for 8xH100 (8x fewer optimizer steps)

**N-gram cache:**
- Hashed n-gram caches are fundamentally broken, not just illegal. Hash collisions inflate all token probabilities — "improvement" tracks collision density, not prediction quality. Confirmed by abaybektursun's PR #886 analysis.

## Key takeaway

At 16MB, architecture is converged. Eval-time AdamW TTT with cosine LR is the remaining legal lever. Everything else is noise.

## Test plan

- [x] All experiments ran to completion
- [x] BPB numbers verified from logs
- [x] Illegal approaches clearly labeled
- [x] Documentation only — no artifacts or code

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1184, opened 2026-03-31
Record: Scylla + Full GPTQ + XSA-all + FA3 — val_bpb 0.9485 (3-seed mean)

icryo:
## Summary
- **val_bpb: 0.9485** (3-seed mean, std 0.0008)
- **~15.6 MB** artifact, 8×H100 SXM, 88ms/step, ~6700 steps
- Combines Scylla tokenizer (PR #1143) with PR #1060 training stack
- **Beats current #1 (PR #1143, 1.0806) by 0.132 BPB (12.2%)**

## 3-Seed Results

| Seed | Sliding BPB (s64) |
|------|--------------------|
| 1337 | 0.9491 |
| 42 | **0.9476** |
| 2025 | 0.9489 |
| **Mean ± Std** | **0.9485 ± 0.0008** |

## Key Innovation
Scylla tokenizer (998 tokens, @simon-marcus PR #1143) + modern training stack:
- Full Hessian GPTQ (Cholesky error compensation)
- XSA on all 11 layers
- Coprime-stride multi-shard loader (194 shards)
- FlashAttention 3 (Hopper native)
- No TTT needed (neutral on this stack)

PR #1143 used the old SOTA base. This submission applies the modern stack to the same tokenizer, yielding 12% better BPB.

## Test plan
- [x] 3-seed verification on 8×H100 SXM (mean 0.9485, std 0.0008)
- [x] All artifacts under 16,000,000 bytes
- [x] All training under 600s
- [x] No TTT
- [x] Tokenizer byte accounting via validated metadata (candidate.meta.npz)

## Credits
- Scylla tokenizer: @simon-marcus (PR #1143)
- Training stack: @resouer (PR #1060), @abaybektursun (PR #549)

PR #511, opened 2026-03-23
Non Record: Add PPM heuristic for test time learning 

AnirudhRahul:
## Summary

Added `records/track_10min_16mb/2026-03-23_10L_Int5MLP_DelayedPPM_K15` with a self-contained snapshot:
- `train_gpt.py`
- `trie_bench.c`
- `submission.json`
- 3 seed logs

This is a **test-time-only** improvement on top of the existing 10L Int5-MLP + BigramHash(10240) model, not a training-time architecture change.

## Method

This submission adds **delayed outside-context-only PPM** at evaluation time.

PPM (Prediction by Partial Matching) is a compression-style backoff n-gram model that boosts a continuation token when the same token prefix has appeared earlier in the document. It prefers longer matches and backs off to shorter ones when confidence is insufficient.

The important constraint is a **2048-token delay**: at position `i`, the PPM bank only contains targets from positions `<= i - 2048`. That means it cannot reuse anything still visible inside the transformer’s current sliding-window context.

So the division of labor is:
- transformer: local context
- delayed PPM: longer-range repeated-sequence signal from outside the local window

## Fixed inference config

- `k = [16, 12, 8, 6]`
- `min_confidence = [1.0, 1.0, 1.0, 0.95]`
- `min_count = [1, 1, 1, 1]`
- `K = 15`
- `delay = 2048`
- `bos_id = 1`

## Per-seed results

| Seed | Baseline val_bpb | Delayed PPM val_bpb | Delta |
|---|---:|---:|---:|
| 42 | 1.14253746 | 1.14125711 | -0.00128035 |
| 1337 | 1.14387335 | 1.14262486 | -0.00124849 |
| 2024 | 1.14257402 | 1.14132993 | -0.00124409 |
| **Mean** | **1.14299494** | **1.14173730** | **-0.00125764** |
| **Std** | **0.00076094** | **0.00076951** | **0.00001979** |

All 3 seeds improved, and all 3 runs remained under the 16MB submission limit.

PR #1272, opened 2026-04-02
Non-record: Comprehensive Negative Results — What Doesn't Work on Strong Models

andrewbaggio1:
## What this is

A collection of things that don't work on well-trained GPTQ'd models.
My coding agents and I ran ~30 experiments over the past two weeks trying
to push below the 1.11 BPP frontier. Most of them failed. This documents
the failures so others don't repeat them.

Updates and extends my earlier negative results PR (#1186).

## Eval-time techniques that don't work on strong models

These all work on weak/undertrained models but provide zero benefit once
your base model is well-trained with Full Hessian GPTQ + sliding window
eval:

| Technique | BPP delta | Why it fails |
|-----------|:---------:|:-------------|
| Properly normalized n-gram (Kneser-Ney, exact trie) | +0.001 to -0.003 | Model is 100x better than n-gram at predicting the correct token. Mixing at any alpha dilutes model confidence. Confirms PR #511 (-0.001) and PR #1145 (-0.003). |
| Online Logit Bias (per-token SGD on logit bias vector) | +0.003 (hurts) | GPTQ'd model is already well-calibrated. No systematic bias to correct. Also takes 1229s (way over eval budget). |
| Prime MLP Adapters (zero-init rank-64, PR #1222 approach) | -0.00009 | PR #1222 got -0.073 but on a 1.50 BPP baseline. Our 1.11 baseline leaves no room — sliding window context already provides everything adapters would learn. |
| Complementary Training (down-weight n-gram-predictable tokens during training) | -0.0004 (noise) | Doesn't change model behavior enough. By the time the model converges, it already knows everything the bigram knows. |
| Score-first chunked TTT (PR #549 approach) | -0.003 | Works but the gain is tiny on GPTQ'd models. PR #1184 also found TTT "neutral" on their stack. |

## The n-gram normalization proof

I built the best possible legal n-gram cache: Kneser-Ney smoothing with
exact trie counts (zero hashing, zero collisions), order 7, full
normalized distribution over all 1024 tokens at every position.

Results on 500K positions:
- Max normalization error: **1.78e-15** (distributions are perfect)
- Zero normalization violations across all positions
- N-gram avg NLL: **5.40** vs model avg NLL: **0.79** (n-gram is 6.8x worse)
- Mixing at ANY alpha hurts on average

The entire 0.09-0.97 BPP improvement from hashed n-gram caches was a
measurement artifact from unnormalized distributions. The real signal from
properly normalized n-grams is 0.001-0.003 BPP, so it's not worth the complexity.

## SLOT violates causal dependence

Detailed in my PR #1240. 100% violation rate across 240 tested pairs.
Self-prediction advantage: +0.24 nats (shared delta), +0.73 nats
(per-sample). Every SLOT-based result on the leaderboard is suspect.

## Scylla tokenizer doesn't help (with correct accounting)

Covered in my other PR. With corrected byte accounting, Scylla gets 1.1289
BPP, the same as SP1024 at 1.1157. The entire sub-1.0 claim was a byte
accounting bug in `candidate.meta.npz`.

## What actually matters

After all these experiments, the model quality is dominated by:
1. **Training data volume** (194+ shards > 80 shards)
2. **Full Hessian GPTQ** (Cholesky + actorder, ~0.005 BPP over naive int6)
3. **Coprime-stride data loader** (batch diversity)
4. **XSA on all layers** (small but consistent gain with coprime loader)

## Files included

- `ngram_test.py` — Kneser-Ney trie with full normalization proof
- `online_logit_bias.py` — Online logit bias implementation + synthetic test
- `correct_meta.npz` — Corrected Scylla byte accounting
- `retokenize_proper.py` — Proper retokenization with official train/val split


andrewbaggio1:
Closing — research documented, moving on to new approaches.

PR #1369, opened 2026-04-05
Non-record: Negative results from gated multi-order hash n-grams

xiayicheng3-code:
## Summary

This PR adds a non-record submission documenting negative results from a gated multi-order hash n-gram line.

The main finding is that a more expressive `2/3/4-gram` gated hash prior did not beat the March 25 BigramHash baseline in the 10min / 16MB regime. The gap showed up before compression and remained after post-training compression. The best legal variant in this line simplified the design to `NGRAM_NUM_HASHES=1,1,1`, reaching:

- post-EMA: `1.1347` val_bpb
- final int6 sliding exact: `1.11959957` val_bpb
- total bytes: `15,937,956`

## Key negative results

- multi-hash hard top-1 routing did not help
- tail-loss emphasis did not improve the roundtrip-to-sliding advantage
- the strongest legal result came from removing same-order hash competition

## Interpretation

One possible interpretation is that hash collisions are not only semantic noise. Because they are deterministic, they may also provide stable pseudo-identity features. If so, aggressive candidate-level semantic gating may remove exactly the stable dirty features that the backbone would otherwise learn to use.

The README also discusses the contrast between Engram-style gated fusion and more additive n-gram fusion styles, and why our results ended up supporting the simpler additive intuition more strongly.

## Practical contributions from this line

Although the model family itself did not win, this line also produced practical experimentation improvements:

- prompt-bank AR calibration
- mild repeat-penalized AR calibration generation
- printed AR calibration outputs for debugging
- next-shard prefetch
- async logging
- faster selective-prune search / bounded undershoot acceptance

## Files added

- README with results, interpretation, and future-work suggestions
- submission.json
- requirements.txt
- train_gpt.py snapshot from the best legal run in this line
- 3 supporting run logs
- results.tsv

### DIFF
--- a/records/track_non_record_16mb/2026-04-04_NegativeResults_GatedMultiOrderHash/README.md
+++ b/records/track_non_record_16mb/2026-04-04_NegativeResults_GatedMultiOrderHash/README.md
@@ -1,0 +1,221 @@
+# Non-Record Submission: Negative Results from Gated Multi-Order Hash N-grams
+
+This non-record submission documents a small line of experiments that tried to replace the simple `BigramHashEmbedding` prior with a more expressive **multi-order hash n-gram** module using scalar gates and, in earlier runs, per-order hard top-1 routing across hash heads.
+
+The main result is negative but, we think, informative:
+
+- the more complex gated hash-n-gram line did **not** beat the March 25 BigramHash SOTA stack in this 10 minute / 16MB regime
+- the main gap showed up **before compression**, not only after GPTQ/pruning
+- once systems regressions were fixed, the best simplified variant got very close in pre-quant quality, but still lost after post-training compression
+
+Our best legal run in this line reached:
+
+- `post_ema val_bpb = 1.1347`
+- `final_int6_sliding_window_exact val_bpb = 1.11959957`
+- `bytes_total = 15,937,956`
+
+This is competitive for a non-record submission, but still behind the March 25 BigramHash reference:
+
+- `post_ema val_bpb = 1.1340`
+- `final_int6_sliding_window_exact val_bpb = 1.11437394`
+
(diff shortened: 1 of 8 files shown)


MatoTeziTanka:
Thanks for publishing this — transparent methodology and the tooling contributions (prompt-bank calibration, bounded pruner acceptance) are genuinely useful.

We tested a related question independently: EngramLite-style gated fusion showed no improvement over baseline BigramHash in our testing. We also ran a separate 4-run experiment on output-level bigram log-prob tables — the neural model learned to suppress the table signal (learnable scale settled at 0.63), confirming additive injection beats structured priors in this regime.

A few observations:

**Batch size confound:** Run 3 (1,1,1 hash, your best) used `batch_tokens=786432` vs `524288` in runs 1-2. That's a ~1.5x batch increase simultaneous with removing hash competition and tail loss. Hard to cleanly attribute the improvement to routing architecture vs training dynamics.

**On the identity-feature hypothesis:** This is the most interesting part. But there's a competing explanation: in a 10-minute budget with -2.0 gate initialization, the gating parameters may simply not have time to learn meaningful routing — underfitting rather than destroying identity features. A quick test: frozen random hash projections. If they perform comparably to learned additive ones, that supports identity. If worse, the problem was training-time failure, not architecture.

**On the overcap run:** Using the 16.2MB run as a capability reference slightly overstates proximity to SOTA — the extra ~200KB of budget is non-trivial at these margins.

The additive-beats-gated conclusion matches our experience. The identity-feature framing is worth testing directly rather than inferring from a single negative result.

*Disclosure: I use Claude Code CLI, Codex CLI, and Gemini Pro as tools in my workflow. Human first, AI-assisted.*

PR #1217, opened 2026-04-01
Non Record: MuonEq-R + Context-Only SLOT + QK_GAIN=5.0 — val_bpb 1.1027 (3-seed mean)

bigbag:
## Summary

**val_bpb: 1.1027** (3-seed mean, std 0.0011) | ≤15.80 MB | 8×H100 SXM | ~88.8ms/step | ~6654 steps

Built on PR #1179 (@dexhunter) with three additions:

- **MuonEq-R** (row-normalization before Newton-Schulz) — from arXiv:2603.28254, ~15 lines
- **QK_GAIN_INIT=5.0** — our hyperparameter sweep across 45 experiments, monotonic gains from 1.5→5.0
- **Context-Only SLOT** — **causal variant** that optimizes delta using only already-scored context tokens (see Legality section)

## 3-Seed Results

| Seed | Context-SLOT BPB | TTT BPB | Steps | ms/step | Artifact |
|------|-----------------|---------|-------|---------|----------|
| 1337 | **1.10166** | 1.11008 | 6660 | 88.8 | 15,795,518 |
| 42 | **1.10378** | 1.11206 | 6650 | 88.9 | 15,793,163 |
| 2024 | **1.10271** | 1.11108 | 6653 | 88.9 | 15,796,779 |
| **Mean** | **1.10272 ± 0.00106** | 1.11107 | 6654 | 88.8 | 15,795,153 |

Beats merged SOTA (PR #1019, 1.1147) by **0.012 BPB** (p ≪ 0.01).

## Improvement Breakdown

| Technique | BPB Impact | Cumulative |
|-----------|-----------|------------|
| PR #1179 base (sliding, no SLOT) | 1.1105 | 1.1105 |
| + MuonEq-R optimizer | -0.001 | ~1.1095 |
| + QK_GAIN=5.0 | -0.001 | ~1.1090 |
| + Context-Only SLOT (8 steps, lr=0.005) | -0.006 | **~1.1027** |

### Training (≤600s on 8×H100)
- Standard transformer training with Parallel Muon optimizer
- **MuonEq-R**: row-normalization before Newton-Schulz orthogonalization (arXiv:2603.28254). Standard optimizer improvement — no rule restricts it.
- **QK_GAIN_INIT=5.0**: hyperparameter choice — no rule restricts it
- Full GPTQ calibration runs within the 600s training budget
- No validation data accessed during training

### Evaluation — Context-Only SLOT (LEGAL, causal by construction)

[...]

PR #1412, opened 2026-04-06
Record: SP8192 + Parallel Residuals + Hessian-Aware SDClip — val_bpb 1.08354 (3-seed mean)

Robby955:
# Record: Parallel Residuals + Hessian-Aware SDClip (3-seed mean 1.08354 BPB)

**val bpb: 1.08354** (3-seed mean, std=0.00050)

Not a record. This is a small 3-seed experiment over [PR #1394](https://github.com/openai/parameter-golf/pull/1394) on my runs, but not enough evidence for a statistical claim, the seed count, and reduction in BPB is too small for confidence. Posting because the changes are zero-cost, reproducible, and may be useful to others trying out different techniques. 

|-|-|-|-|-|-|
| 1337 | 5178 | 1.08765 | 1.09959 | **1.08301** | 15,976,275 |
| 42 | 5180 | 1.08816 | 1.10013 | **1.08363** | 15,978,439 |
| 3141 | 5182 | 1.08872 | 1.10044 | **1.08399** | 15,979,649 |
| **Mean** | | 1.08818 | 1.10005 | **1.08354** | 15,978,121 |

## Changes

Three zero-cost modifications on top of [PR #1394](https://github.com/openai/parameter-golf/pull/1394), adding zero extra parameters or bytes:

### 1. Parallel Residuals (Layers 7+)

GPT-J style parallel attention+MLP ([Wang & Komatsuzaki, 2021](https://github.com/kingoflolz/mesh-transformer-jax)) for the last 4 layers. Both attention and MLP read from the same input and their outputs are added in parallel:

```
# Parallel (layers 7-10):
x_out = x + attn_scale * Attn(norm(x)) + mlp_scale * MLP(norm(x))

# Sequential (layers 0-6, unchanged):
h = x + attn_scale * Attn(norm(x))
x_out = h + mlp_scale * MLP(norm(h))
```

I expected parallel residuals to reduce interference between attention and MLP during GPTQ calibration. Pre-quant BPB barely moved, but the quantization gap tightened across all 3 seeds, which made this the most useful change in practice.

### 2. Hessian-Aware SDClip

I used GPTQ's existing Hessian diagonal as a cheap importance signal to slightly modulate SDClip thresholds by row:

$$c_i = k \cdot \sigma_i \cdot [1 + \lambda(r_i - 1)], \quad \lambda = 0.175$$

where $\sigma_i$ is the standard deviation of row $i$ and $r_i$ is the row importance derived from Hessian-weighted magnitude. The effect is small but directionally useful at $\lambda = 0.175$; higher $\lambda$ hurt compression. I initially used $\lambda = 0.30$ but found $\lambda = 0.175$ is consistently better across seeds — both lower BPB and smaller artifact. Higher $\lambda$ reduces rounding error but increases entropy, which makes Brotli compression less effective.

### 3. Progressive Recurrence

Depth recurrence split into two phases: first loop enabled at 50% of training, second at 65%. The split points were not optimized — 50% matches the original and 65% was a single manual choice. Enabling both loops at once causes a sharper loss spike; splitting gives the model time to adapt to each additional pass before adding the next.

## Hessian Analysis (Cross-Seed)

Hessian diagnostics from 3 seeds, 67 matrices each:

- **Group-level traces** (early/loop/mid/late blocks): $r=0.997$ across seeds
- **Per-matrix traces**: $r=0.994$
- **Per-row importance**: $r=0.12$ (noise)

Importance hierarchy: early blocks (30x trace of late blocks) >> loop >> mid >> late. Per-row importance is too noisy to be a reliable signal, but group-level traces are very stable across seeds. This suggests per-group clip allocation could be a useful direction.

## Future Directions

Several ideas I'd like to explore with more compute time:

- **Per-group clip allocation**: Non-uniform $k$ across layer groups, using the stable group-level trace hierarchy as a guide.
- **Output-Hessian weighting**: Using backward-pass gradients for output-side row importance rather than input-side alone.
- **More seeds**: 3 seeds is not enough for strong statistical claims. I'd want 5+ to be confident about the gap vs PR #1394.
- **YAQA**: I like the idea of the paper ([arXiv:2505.22988](https://arxiv.org/abs/2505.22988)), but I couldn't get a working backward pass for it. I think maybe it could be adapted for the parameter golf problem in an interesting way. I also like the math in Mousse ([arXiv:2603.09697](https://arxiv.org/abs/2603.09697)), but exploiting curvature in small LMs seems tough.

## Run Command

```bash
HESSIAN_CLIP_LAMBDA=0.175 LOOP_PHASE2_AT=0.65 PARALLEL_RESIDUAL_START=7 SEED=1337 \
torchrun --standalone --nproc_per_node=8 train_gpt_sweep.py
```

## Requirements

Flash Attention 3 (Hopper) required. SP8192 BPE tokenizer trained on FineWeb 10B (sentencepiece BPE, 8192 vocab).

```bash
pip install torch --index-url https://download.pytorch.org/whl/cu130
pip install --no-cache-dir \
  "https://download.pytorch.org/whl/cu130/flash_attn_3-3.0.0-cp39-abi3-manylinux_2_28_x86_64.whl"
pip install -r requirements.txt
```

## Compliance (Track A — Fixed Predictor)

- No TTT, SLOT, n-gram cache, or eval-time adaptation
- GPTQ calibration within training budget
- Standard autoregressive sliding-window eval (stride=64)

## Credits

Learned from and inspired by [PR #1394](https://github.com/openai/parameter-golf/pull/1394) (@clarkkev) — SDClip, depth recurrence, and GPTQ embedding quantization ideas. Parallel residuals from GPT-J ([Wang & Komatsuzaki, 2021](https://github.com/kingoflolz/mesh-transformer-jax)). Additional credits: PR #1204 (@msisovic, depth recurrence), PR #1217 (@bigbag, MuonEq-R), PR #1019 (@abaybektursun, previous SOTA).

### DIFF
--- a/records/track_10min_16mb/2026-04-06_SP8192_HessianSDClip_ProgressiveRecurrence/README.md
+++ b/records/track_10min_16mb/2026-04-06_SP8192_HessianSDClip_ProgressiveRecurrence/README.md
@@ -34,9 +34,9 @@ I expected parallel residuals to reduce interference between attention and MLP d
 
 I used GPTQ's existing Hessian diagonal as a cheap importance signal to slightly modulate SDClip thresholds by row:
 
-$$c_i = k \cdot \text{std}(\text{row}_i) \cdot [1 + \lambda(\text{row\_importance}_i - 1)], \quad \lambda = 0.175$$
+$$c_i = k \cdot \sigma_i \cdot [1 + \lambda(r_i - 1)], \quad \lambda = 0.175$$
 
-where row importance is derived from Hessian-weighted magnitude. The effect is small but directionally useful at $\lambda = 0.175$; higher $\lambda$ hurt compression. I initially used $\lambda = 0.30$ but found $\lambda = 0.175$ is consistently better across seeds — both lower BPB and smaller artifact. Higher $\lambda$ reduces rounding error but increases entropy, which makes Brotli compression less effective.
+where $\sigma_i$ is the standard deviation of row $i$ and $r_i$ is the row importance derived from Hessian-weighted magnitude. The effect is small but directionally useful at $\lambda = 0.175$; higher $\lambda$ hurt compression. I initially used $\lambda = 0.30$ but found $\lambda = 0.175$ is consistently better across seeds — both lower BPB and smaller artifact. Higher $\lambda$ reduces rounding error but increases entropy, which makes Brotli compression less effective.
 
 ### 3. Progressive Recurrence
 
--- a/records/track_10min_16mb/2026-04-06_SP8192_HessianSDClip_ProgressiveRecurrence/README.md
+++ b/records/track_10min_16mb/2026-04-06_SP8192_HessianSDClip_ProgressiveRecurrence/README.md
@@ -1,0 +1,91 @@
+# Non-record: Parallel Residuals + Hessian-Aware SDClip (3-seed mean 1.08354 BPB)
+
+**val bpb: 1.08354** (3-seed mean, std=0.00050)
+
+Not a record. This is a small 3-seed experiment over [PR #1394](https://github.com/openai/parameter-golf/pull/1394) on my runs, but not enough evidence for a statistical claim — the seed count is too small for confidence. Posting because the changes are zero-cost, reproducible, and may be useful to others trying out different techniques. 
+
+| Seed | Steps | Pre-quant BPB | Post-quant BPB | **Sliding BPB** | Artifact |
+|-|-|-|-|-|-|
+| 1337 | 5178 | 1.08765 | 1.09959 | **1.08301** | 15,976,275 |
+| 42 | 5180 | 1.08816 | 1.10013 | **1.08363** | 15,978,439 |
+| 3141 | 5182 | 1.08872 | 1.10044 | **1.08399** | 15,979,649 |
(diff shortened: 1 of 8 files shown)


Robby955:
## Artifact Size Clarification

The weight blobs reported in the table above (`artifact_bytes`) are just the compressed model weights. The full submission artifact = weights + `train_gpt.py` code.

When I originally ran this, `train_gpt.py` used `exec(open("train_gpt_decode.py").read())` which meant the effective code size was the full readable script (~78KB), pushing the total to ~16.05MB — slightly over the 16MB limit.

Following the same LZMA packing technique used in the base PR #1394 (compressing the Python source with LZMA and base85-encoding it into a 2-line `train_gpt.py`), the code shrinks from ~78KB to ~23KB:

| Component | Before | After (LZMA-packed) |
|-----------|--------|---------------------|
| Weights (seed 1337) | 15,976,275 | 15,976,275 (unchanged) |
| Code (`train_gpt.py`) | 78,688 | 23,095 |
| **Total** | **16,054,963** (over) | **15,999,370** (under by 630) |

No weights, BPB results, or model architecture changed — this is purely a code packaging fix. The readable `train_gpt_decode.py` (included in the PR as `train_gpt_human.py` equivalent) is unchanged.

PR #1461, opened 2026-04-08
Record Submission: HDC_DSV_Hadamard_Spiral_val_bpb: 0.4067 

viasky657:
This is an exotic architecture with primary learning mechanisms involving directional semantic vectors, golden axis shift with a Sylvester Hadamard Matrix, and a code cookbook table. It achieves 0.411 val_bpb on the 3 independent runs on the H100x8 pod within the time limit. (After the update with the golden axis shift and bug fixes, it now runs at 0.4067 BPB.)


MatoTeziTanka:
Thanks for the detailed writeup — the acronym stack (HDC = Hyperdimensional Computing, DSV = Directional Semantic Vectors, 1-Step Grad, Sylvester/Hadamard symmetry, radial) describes a genuinely unusual pipeline, and it's clear from the README and the ~10k-line diff that a lot of care went into the multi-phase build (G[p] rolling hash → NMF frequency table → Fibonacci-hash XOR bundles → suffix grammar → LZMA9 artifact). Reviewing as a community member before mods weigh in on the claimed **0.4118 BPB** (vs. current SOTA **1.0810** and naive baseline **1.2244**). A few questions that would help me and anyone else reading the PR evaluate the claim.

### 1. Is the eval `p_correct` a normalized probability distribution?

In [`_hash_grad_train.py:919–947`](records/track_10min_16mb/2026-04-07_HDC_1_Step_Grad_DSV_Radial_Slyvester_Hadamard_Matrix_Symmetry/2026-04-07_HDC_1_Step_Grad_DSV_Radial_Slyvester_Hadamard_Matrix_Symmetry/_hash_grad_train.py) the miss-path predictor is:

```python
sv = sem_fwd[prev_t]           # XOR-bundle for the previous token
tv = codebook[m_tgt]            # <-- the TARGET token's hypervector
xv = sv ^ tv
pc = np.unpackbits(xv.view(np.uint8), axis=1).sum(axis=1)
conf = np.abs(pc - half) / half
p_sem = np.clip(0.5 + 0.49 * conf, 1e-30, 0.99)
```

and then `total_bits += -np.log2(p_sem).sum()`. The same pattern is used in the collision path (lines 891–897) and in the skip-bigram lag blend (lines 931–947). In every case the confidence is computed directly against `codebook[target]`, so `p_sem` is a one-number-per-position similarity score mapped into `[0.5, 0.99]`, not a component of a softmax over the 1024-token vocabulary.

The competition metric requires
`BPB = Σ(-log₂ P(y_t | context)) / Σ(bytes(y_t))`
where `P(·|context)` is a probability distribution over the vocabulary (so it sums to 1 across all possible targets). The reference `train_gpt.py:eval_val` enforces this by using `F.cross_entropy` on a normalized softmax.

Could you point me to where this path normalizes across the vocabulary? As written, `Σ_v p_sem(v | prev_t)` is not 1 — for any given `prev_t` you'd get a different `p_sem` for each candidate target, and they aren't forced to sum. Is there a step I'm missing that converts these similarities into a proper distribution, or is the `-log₂(0.5 + 0.49·conf)` being used directly as if it were `-log₂ P(y_t|x)`?

If it's the latter, the `val_loss = 0.6868 nats/token` number doesn't correspond to a language-model cross-entropy in the sense the leaderboard uses, and the 0.4118 BPB would not be comparable to the other leaderboard entries.

### 2. Does the Tier-1 (NMF) path dominate, and can you show its contribution separately?

The README explicitly says the NMF is run with `nmf_max_iter=1`, that the KL loss stays near `ln(vocab_size)`, and that "the DSV is the dominant signal path." If that's true, then nearly all of the 0.9908 bits/token is coming from the DSV/skip-bigram similarity path described in (1). Could you add a per-path breakdown to the log — how many tokens went through the filled-bucket NMF path vs. the fingerprint-collision path vs. the miss path, and the average `-log₂ p` on each? That would let a reader confirm where the bits are actually coming from.

### 3. Byte-count sanity check

I reproduced the SentencePiece LUT build against `fineweb_1024_bpe.model` on the first 5M val tokens on a CPU box and get `avg bytes/token = 2.4055`, matching the `2.4062` in the `[HashGrad BPB audit]` block. So the denominator of the BPB formula looks correct — no complaint on byte counting. The question in (1) is strictly about the numerator.

### 4. CPU gauntlet

I ran the CPU pre-flight gauntlet we've been using on record PRs (`CUDA_VISIBLE_DEVICES=""`, stubbed flash_attn / cutlass_evt, import-only + `Hyperparameters` discovery). The submitted `train_gpt.py` imports cleanly under the stubs but has no `Hyperparameters` class, so the standard pre-flight stops at that point — the record uses a `--hash_grad` argparse entry point instead of the baseline shape. That's not a compliance blocker on its own (a record is allowed to replace the whole script), but it does mean this submission can't be exercised through the same smoke path as the baseline-shaped PRs. If you could add a minimal CPU path (even one that asserts `torch.cuda.is_available()` early), that would make independent verification much easier.

### 5. Reproducibility and 3-run logs

The three logs (`train_20260408T011446.log`, `_012041.log`, `_012629.log`) all report `BPB: 0.4118` with `σ = 0.0000` across seeds `42, 7, 1337`. That's suspicious for a real 3-seed average — on the order of the leaderboard's other record submissions, run-to-run variance shows up in the 4th decimal. Given that all three runs go through the same `hash_grad_bpb` path, a zero-variance result is consistent with the score being deterministic given the seeds, but it's also consistent with the scoring path being insensitive to the model parameters (which would happen if `p_sem` is dominated by a data-independent term). Seeing per-path bit contributions (question 2) would resolve that.

### 6. A concrete reproducibility ask

Could you run one additional evaluation that replaces the `hash_grad_bpb` waterfall with a call to the stock `eval_val()` from the root `train_gpt.py`, using `embed @ W_out` as the logits (zeros for unseen buckets)? That's the most direct apples-to-apples comparison to the leaderboard baseline and would immediately show whether the 0.4118 number survives when scored against a real softmax. The logits are already produced in the filled-bucket branch, so it should be a ~20-line change.

---

I want to reiterate that I'm reviewing this in good faith — HDC + Sylvester/Hadamard + XOR-bundle predictors is a legitimately interesting corner of the design space, and if the score holds up it would be a massive result. But a ~0.66 BPB jump over the current SOTA on a pipeline built from bigram-level structure is the kind of claim that needs the eval path to be airtight, and right now I can't convince myself that the `p_sem = 0.5 + 0.49·conf(target)` step produces a proper probability distribution over the vocabulary. Happy to be shown I'm wrong.

---
*Reviewed by [@MatoTeziTanka](https://github.com/MatoTeziTanka) — [The Agora](https://matotezitanka.github.io/parameter-golf). CPU gauntlet (CT2038 proteus-engine, 2026-04-11): FAIL (no `Hyperparameters` class — record uses custom `--hash_grad` entry point). AI tooling: review drafted with Claude Code (Sonnet/Opus) using an internal review template; all citations, file paths, and compliance audits were verified against the PR's actual code at SHA `d3087740c23b164be4c52cb587bf5efabefe689c`.*

viasky657:
Thank you for the great detailed feedback!                                                                                                                                       For 4., I haven't tested it on the CPU preflight check for the final results so it might not produce the same expected results as the GPU. I have only tested the GPU which can be a bit varied in behavior due to how floats are sometimes calculated in comparison to the slightly different bytes calculation that the HDC uses for massive parallelism on CPU. I will fix the code so that they both use a similar calculation for fairness and accuracy.                                                                                                       For 2., I recently explored the NFM more thoroughly and it seems to not contribute at all to the final results and that only the DSV really contributes so it can be completely cut from the pipeline for clarity.                                                                                                                    I will definitely look into 5. and 6. as well to be sure that the calculations matches what is expected so it is a fair comparison against the transformer baseline for the competition.

Qeuph:
Hmm

viasky657:
I updated the architecture to use the full training dataset for all runs. I also updated the 3 seed runs to run independently; they will likely still have the same output or close to it because the architecture is deterministic with very little noise or accuracy loss. This is partially due to the new golden axis shift which allows for more lags and orthogonality between rotations and positions in the token sequence. I clarified what each component contributes to the final bits per byte outcome in greater detail in the readme. I also fixed a bug where the entire evaluation dataset was not used from before. The context window was extended for this new model version. In addition, I removed a dead backwards pass from before since only the forward pass contributed meaningfully to the model's bits per bytes level. The code book contributes to the model's architecture a lot more than the DSV alone so I kept both. The DSV primarily corrects rare tokens in the code book while the code book has semantic information organized innately using the Sylvester Hadamard and pop count [+1 (positive relationship), 0 (neutral), -1 (negative relationship)]. I explain more of my findings in more detail in the readme. I am planning on implementing the suggestions for evaluation in the next iteration. I got the new BPB down from 0.4118 to approximately BPB 0.4067 for this next iteration due to adding the golden axis in and the minor fixes in the architecture that reduced accuracy.

viasky657:
I added in clear labels for Bits per Byte and learning tracking especially for the directional semantic vector to show how the model learns. It primarily learns using similarity and pop count in place of frequency and softmax args in order to have controlled randomness. The pop count has correlations and sharper relationships between subatomic (bits) and atomic (bytes) components of the tokens to better understand the structure of the dataset and learn the vocabulary through similarity bundling and search. The gradient hash handles the common words and structures for learning while the DSV handles the rest. This model gets its primary learning signal from the DSV so I did include a pure hash gradient bits per byte but the architecture would fail without having the similarity structure.  I also fixed the issue with the CPU preflight check not working with some of the imports so it should work now but the model can only be trained using GPUS.

PR #1519, opened 2026-04-10
BPB-weighted training loss: align training objective with eval metric

elliottdehn:
## Summary

Weight each token's cross-entropy loss by the number of UTF-8 bytes it encodes, directly aligning the training objective with the BPB evaluation metric. Three lines changed in `train_gpt.py`.

## The Change

```python
# Standard (uniform token weighting):
return F.cross_entropy(logits.float(), targets, reduction="mean")

# BPB-weighted (bytes-per-token weighting):
if self.training and hasattr(self, '_byte_weights'):
    per_token_loss = F.cross_entropy(logits.float(), targets, reduction="none")
    w = self._byte_weights[targets]
    return (per_token_loss * w).sum() / w.sum()
return F.cross_entropy(logits.float(), targets, reduction="mean")
```

`_byte_weights` is `base_bytes_lut.float().clamp(min=1.0)`, registered as a buffer after model creation. `base_bytes_lut` already exists in the codebase for BPB evaluation — we just reuse it during training.

## Why It Works

The eval metric is bits per byte, but standard CE weights all tokens equally. A token encoding "the " (4 bytes) contributes 4x to BPB compared to a single-byte token, but gets equal weight in training. BPB-weighting shifts gradient toward multi-byte tokens that matter most for the eval metric.

Works specifically because the 1024-token SentencePiece vocab has gentle byte-length variance (1-8x). Verified it does NOT work with large vocabularies (GPT-2 50K) where extreme byte lengths destabilize training.

## Preliminary Results (2x RTX 5090, SDPA fallback)

Compared against unmodified baseline on same hardware, same seed:

| Step | Baseline val_bpb | BPB-weighted val_bpb | Delta |
|------|-----------------|---------------------|-------|
| 500 | 1.3841 | 1.3763 | -0.0078 |
| 1000 | 1.3076 | 1.2961 | -0.0115 |
| 1500 | 1.2823 | 1.2693 | -0.0130 |
| 2000 | 1.2512 | 1.2366 | -0.0146 |
| 2500 | 1.2375 | 1.2217 | -0.0158 |
| 7000 | — | 1.1155 | -0.0194 |

Post-EMA: **1.1146** vs current record's post-EMA 1.1340. Delta: **-0.0194 bpb**.

Gap widens monotonically through training.

## Status

**Preliminary / non-record submission.** Results are on 2x RTX 5090, not the official 8xH100 environment. Awaiting compute grant for:
- 3-seed runs on 8xH100 for statistical significance (p < 0.01)
- Full pipeline (EMA + SWA + GPTQ + sliding window eval)
- Official submission with records folder, logs, and submission.json

Will update this PR with official results once H100 runs are complete.

## Properties

- **Zero extra parameters**
- **Zero extra compute** (one index lookup per token)
- **Fully composable**: stacks on top of any architecture, optimizer, or quantization change
- **Not hardware-dependent**: delta should transfer across hardware

### DIFF
--- a/train_gpt.py
+++ b/train_gpt.py
@@ -721,6 +721,11 @@ class GPT(nn.Module):
                 raise RuntimeError("lm_head is required when tie_embeddings=False")
             logits_proj = self.lm_head(x)
         logits = self.logit_softcap * torch.tanh(logits_proj / self.logit_softcap)
+        # BPB-weighted loss: weight tokens by byte count to align with eval metric
+        if self.training and hasattr(self, '_byte_weights'):
+            per_token_loss = F.cross_entropy(logits.float(), targets, reduction="none")
+            w = self._byte_weights[targets]
+            return (per_token_loss * w).sum() / w.sum()
         return F.cross_entropy(logits.float(), targets, reduction="mean")
 
 
@@ -836,6 +841,9 @@ def main() -> None:
         rope_base=args.rope_base,
         qk_gain_init=args.qk_gain_init,
     ).to(device).bfloat16()
+    # BPB-weighted loss: register byte count per token as buffer
+    base_model.register_buffer('_byte_weights', base_bytes_lut.float().clamp(min=1.0))
+    log0(f"bpb_weighted_loss:enabled mean_bytes_per_token:{base_bytes_lut.float().clamp(min=1.0).mean():.2f}")
     for module in base_model.modules():
         if isinstance(module, CastedLinear):
             module.float()
@@ -1065,15 +1073,17 @@ def main() -> None:
     # Save the raw state (useful for debugging/loading in PyTorch directly), then always produce
     # the compressed int8+zlib artifact and validate the round-tripped weights.
 
+    # Filter training-only buffers from saved state
+    save_sd = {k: v for k, v in base_model.state_dict().items() if k != '_byte_weights'}
     if master_process:
-        torch.save(base_model.state_dict(), "final_model.pt")
+        torch.save(save_sd, "final_model.pt")
         model_bytes = os.path.getsize("final_model.pt")
         code_bytes = len(code.encode("utf-8"))
         log0(f"Serialized model: {model_bytes} bytes")
         log0(f"Code size: {code_bytes} bytes")
         log0(f"Total submission size: {model_bytes + code_bytes} bytes")
 
-    quant_obj, quant_stats = quantize_state_dict_int8(base_model.state_dict())
+    quant_obj, quant_stats = quantize_state_dict_int8(save_sd)
     quant_buf = io.BytesIO()
     torch.save(quant_obj, quant_buf)
     quant_raw = quant_buf.getvalue()
@@ -1096,7 +1106,7 @@ def main() -> None:
     with open("final_model.int8.ptz", "rb") as f:
         quant_blob_disk = f.read()
     quant_state = torch.load(io.BytesIO(zlib.decompress(quant_blob_disk)), map_location="cpu")
-    base_model.load_state_dict(dequantize_state_dict_int8(quant_state), strict=True)
+    base_model.load_state_dict(dequantize_state_dict_int8(quant_state), strict=False)
     torch.cuda.synchronize()
     t_qeval = time.perf_counter()
     q_val_loss, q_val_bpb = eval_val(


definenoob:
Tagging myself, as @definenoob is my alt account

definenoob:
Apologies, I had an out of date repo. Ignore this PR.

MatoTeziTanka:
Thanks for writing this up @elliottdehn — a three-line training-time loss reweight tied to the eval metric is exactly the kind of clean, composable technique proposal worth landing on its own merits, even before an official 8xH100 run.

A few notes from reading the diff at `0000334ffb9a248e25867c64af1e629c5bcdd27e`:

## What the change actually is

Three edits to `train_gpt.py`:

1. `GPT.forward` (line ~721): when `self.training and hasattr(self, '_byte_weights')`, replace `F.cross_entropy(..., reduction="mean")` with a byte-weighted mean, `(per_token_loss * w).sum() / w.sum()`, where `w = self._byte_weights[targets]`.
2. After model construction (line ~841): `base_model.register_buffer('_byte_weights', base_bytes_lut.float().clamp(min=1.0))`. The `base_bytes_lut` already exists in the script for BPB eval, so this is a pure reuse of an existing per-token byte count.
3. Serialization (line ~1065): `save_sd` drops `_byte_weights` from the state dict before `torch.save` and `quantize_state_dict_int8`, and the final roundtrip `load_state_dict(..., strict=False)` tolerates the missing buffer.

`model.eval()` is called inside `eval_val`, so the `self.training` gate makes `eval_val` fall through to the standard unweighted CE. I walked the eval path and nothing else changed — `base_bytes_lut`, `has_leading_space_lut`, `is_boundary_token_lut`, the byte accounting, and the `val_bpb = bits_per_token * tokens_per_byte` computation are all untouched. The eval metric is the same one every other submission is graded against, which makes this a legitimate training-side technique: "Training-side modifications are evaluated by the val BPB they produce, not by the loss landscape — the test is whether the model achieves a lower stride-64 sliding-window BPB on the canonical eval."

## On the BPB numbers

Just to clear up the numbers for anyone skimming (I had to squint at them too): the `-0.0194` in the table is the **delta** between baseline and BPB-weighted at step 7000, not a final val BPB. The post-EMA number you actually report is **1.1146** vs baseline **1.1340** on 2xRTX 5090 with SDPA fallback. That's a preliminary non-record claim, clearly framed as such, waiting on 8xH100 for statistical significance — which is the honest way to file this.

A couple of questions on that front:

1. The table shows the gap **widening monotonically** from -0.0078 at step 500 to -0.0158 at step 2500 and -0.0194 by step 7000. Does the gap eventually saturate, or does it keep opening? That would be useful to know before running the 3-seed H100s, since a widening gap at 2xRTX might narrow on the longer wallclock available at 8xH100 (where baseline reaches lower-entropy regimes faster).
2. You mention BPB-weighting **destabilizes training at GPT-2 50K vocab** but works at SP1024. Do you have any intuition for where the vocab-size cliff sits? A quick sentence in the PR body on why 1-8x byte variance is fine but larger vocab ranges blow up would help anyone thinking of porting this to a different tokenizer.
3. Minor: the `mean_bytes_per_token` is logged at init, but the effective gradient scale of the weighted loss is `sum(w) / len(w)` times the unweighted loss gradient only in expectation. Is there any implicit LR change you noticed, or did the per-group LRs work unchanged?

## Compliance audit

- Loss change is gated on `self.training`, so eval uses unmodified CE. [clean]
- `_byte_weights` is a registered **buffer**, not a parameter — doesn't enter the optimizer param groups and doesn't count toward the param budget. [clean]
- Filtered out of `save_sd` before serialization, so the 16MB artifact budget is unaffected. [clean]
- `base_bytes_lut` is already computed from the SentencePiece model via `build_sentencepiece_luts` for the existing BPB metric; reusing it as training weights doesn't introduce any new data dependency. [clean]
- `strict=True` → `strict=False` on the post-roundtrip `load_state_dict` is the only side effect — necessary because the reinstantiated `base_model` has `_byte_weights` but the quantized payload does not. A tighter alternative would be `del base_model._byte_weights` before the roundtrip load and keep `strict=True`, which I'd marginally prefer because `strict=False` silently tolerates future missing keys too. Not a blocker.

## CPU gauntlet (CT2038 proteus-engine, 2026-04-11)

```
[PASS] Import (0.6s)
[PASS] Hyperparameters: dim=512, layers=9, heads=8, vocab=1024
[PASS] Model: 17,059,912 params (6.3s)
[PASS] Forward pass: loss=6.9354 (47.3s)
[INFO] Est. 1xH100: 154.9ms/step, 3874 steps in 10min
[INFO] Weights: 55 matrices, avg_kurtosis=-1.77, avg_max/std=1.2
```

Import clean, param count identical to baseline (17,059,912), random-init forward loss 6.9354 — all consistent with "no architectural change, training-side loss reweight only." The wrapper's artifact check reports FAIL at 68MB because the CPU pre-flight dumps raw fp32 state without running the int8+zlib path — this is the standard baseline behavior on the CPU gauntlet and not a PR-specific regression.

## Verdict

This looks like a clean, small, composable technique contribution. Happy to see the 3-seed 8xH100 numbers land; the preliminary delta is large enough that even a significant haircut on the official run would still make this worth keeping as an orthogonal training-side lever. The framing of aligning the training objective with the eval metric is the right way to motivate it, and the three-line implementation (plus one strict-mode relaxation) is minimally invasive.

Non-blocking suggestions if you do a follow-up push:
- Swap `strict=False` for an explicit `del base_model._byte_weights` before the roundtrip load.
- One sentence in the PR body clarifying that the `-0.0194` is a step-7000 delta, not a final val BPB, for readers who just glance at the table.
- If it's cheap, a quick ablation with `clamp(min=1.0)` replaced by identity (i.e. respecting the zero-byte control tokens) — curious whether the clamp is load-bearing or just defensive.

---
*Reviewed by [@MatoTeziTanka](https://github.com/MatoTeziTanka) — [The Agora](https://matotezitanka.github.io/parameter-golf). CPU gauntlet (CT2038 proteus-engine, 2026-04-11): 17,059,912 params, forward loss 6.9354, import+forward clean. AI tooling: review drafted with Claude Code (Sonnet/Opus) using an internal review template; all citations, file paths, and compliance audits were verified against the PR's actual code at SHA `0000334ffb9a248e25867c64af1e629c5bcdd27e`.*

MatoTeziTanka:
### Community Review — BPB-weighted training loss: align training objective with eval metric

PR #1519 modifies `train_gpt.py` only. The full 1136-line file was reviewed.

**N-gram / XOR family bug check — CLEAR**
No `ctx_hash`, `full_key`, `primes[k]`, or target-XOR-into-hash-key patterns found. The only XOR-adjacent mention is at line 943 (a comment about "warmup primes the compiled…" path — unrelated). No BigramHash or n-gram lookup table of any kind is present.

**TTT / Pre-Quant TTT check — CLEAR**
No test-time training, no optimizer step on val_tokens, no multi-epoch AdamW on val data. `val_tokens` is used exclusively in `eval_val()` which runs fully under `torch.inference_mode()` (lines 250–278) with no gradient tracking and no optimizer calls. No `is_last_chunk` guard, no chunk-wise TTT loop, no score-first pattern — none needed because TTT is simply absent.

**Scored-region SLOT check — CLEAR**
No masking + optimize + score-same-region pattern.

**What the PR actually does:**
Two legitimate training-quality changes:
1. **BPB-weighted loss** (lines 725–728, 844–846): During training, cross-entropy is weighted per-token by byte count (`_byte_weights` buffer derived from `base_bytes_lut`). This aligns the training objective more closely with the BPB eval metric. Weights are derived from the tokenizer's byte-length lookup table, not from val_tokens. Fully legal.
2. **Muon momentum warmup** (lines 82–83, 1029–1032): `MUON_MOMENTUM_WARMUP_START` ramps from 0.85 to `muon_momentum` over `MUON_MOMENTUM_WARMUP_STEPS` iterations. Standard optimizer scheduling tweak. Legal.

No rule violations found. The submission is a pure neural training improvement.

**Verdict:** LOOKS CLEAN.

---

PR #1523, opened 2026-04-10
Record: SP8192 + Triple Recurrence + Banking + Fused MLP + Muon 0.97 — val_bpb 1.0778 (3-seed mean)

EthanYangTW:
## Record: SP8192 + Triple Recurrence + Banking + Fused MLP + Muon 0.97

**val_bpb = 1.0778** (3-seed mean, std 0.0008) | **~15.99 MB** | 8xH100 SXM

### 3-Seed Results

| Seed | Sliding BPP | **TTT BPP** |
|------|-------------|-------------|
| 1337 | 1.0786 | **1.0771** |
| 42 | 1.0792 | **1.0776** |
| 2024 | 1.0798 | **1.0787** |
| **Mean** | **1.0792** | **1.0778** |

Merged SOTA (PR #1493): **1.0810 BPP**. Delta: **-0.0032 BPP**.

### Key Technical Contributions

1. **Parameter Banking + Parallel Muon** — 66 matrices → 4 contiguous banks, batched Newton-Schulz (15x faster optimizer step), +3.8% throughput
2. **Fused MLP Triton TMA Kernel** — fc→LeakyReLU(0.5)→square in one kernel, +2% throughput. Combined with banking: +5.2% total
3. **Muon Momentum 0.97** — reduced from 0.99, -0.0004 BPP
4. **Triple Depth Recurrence** — 17 virtual layers from 11 physical (loop 3,4,5 x3, enable at 35%)
5. **Eval-Time Hash Embedding** — zero-init 16384×512 bigram hash, trained in TTT loop
6. **TTT LR=0.01** — optimized from default 0.005, -0.0003 BPP

### Compliance (Track B)

- Score-first TTT: every chunk scored under no_grad() before gradient update
- No SLOT, no pre-quant TTT, no n-gram caches, no Tap-In
- All 4 conditions from Issue #1017 satisfied

### Reproduction

```bash
pip install brotli sentencepiece
SEED=1337 TTT_ENABLED=1 HASH_EMBED_ENABLED=1 TTT_LR=0.01 MUON_MOMENTUM=0.97 \
  torchrun --standalone --nproc_per_node=8 train_gpt.py
```

### Credits

PR #1420 @abaybektursun, PR #1394 @clarkkev, PR #1471 @X-Abhishek-X, PR #1477 @aryanbhosale, PR #1460 @resouer, PR #399 @abaybektursun, PR #1514 @dexhunter

PR #1525, opened 2026-04-10
Non-Record 16 MB Track : Hybrid XSA-SSM 

Jash-Vora:
**val_bpb: 1.255216** (post-quantization) | **1.2146 pre-quantization** | 14,915,938 bytes | Stopped at step 3723 (600s wallclock cap)

Built on PR #549. 8×H100 SXM | PyTorch 2.4.0-py3.11-cuda12.4.1

---

## SSM Hybrid Layer

The core novelty of this submission is replacing the middle transformer layer (layer 5, i.e. `num_layers // 2`) with a [Mamba](https://github.com/state-spaces/mamba) SSM block instead of a standard attention block:

```python
mamba = Mamba(d_model=512, d_state=64, d_conv=4, expand=2)
```

It is inserted as a residual contribution:

```python
if isinstance(block, Mamba):
    x_out = x + run_mamba(block, x.to(torch.bfloat16)).to(x.dtype)
```

Placing it at the midpoint of the U-Net-style encoder–decoder stack lets the SSM act as a bottleneck that compresses long-range sequence state before the decoder layers unpack it. The surrounding transformer layers handle local attention patterns as usual.

The Mamba block's internal parameters (`A_log`, `dt_bias`, `D`) are routed to the AdamW scalar optimizer rather than Muon, since they don't benefit from Newton-Schulz orthogonalization. The SSM block contributes ~1.84M of the model's ~25.95M total parameters.

`@torch._dynamo.disable` is applied to the Mamba forward call to avoid TorchDynamo tracing issues with the SSM's custom CUDA kernels.

---

## Results

| Metric | Value |
|---|---|
| Pre-quantization val_bpb | 1.2146 |
| Post-quantization val_bpb | 1.255216 |
| val_loss (nats) | 2.119375 |
| Steps completed | 3723 / 20000 |
| Wallclock | 600s |
| Artifact size | 14,915,938 bytes |

### DIFF
--- a/records/track_non_record_16mb/2026-04-10_Hybrid-SSM-XSA/README.md
+++ b/records/track_non_record_16mb/2026-04-10_Hybrid-SSM-XSA/README.md
@@ -1,0 +1,41 @@
+# Hybrid XSA-SSM — Non-Record 16 MB Track
+
+**val_bpb: 1.255216** (post-quantization) | **1.2146 pre-quantization** | 14,915,938 bytes | Stopped at step 3723 (600s wallclock cap)
+
+Built on PR #549. 8×H100 SXM | PyTorch 2.4.0-py3.11-cuda12.4.1
+
+---
+
+## SSM Hybrid Layer
+
+The core novelty of this submission is replacing the middle transformer layer (layer 5, i.e. `num_layers // 2`) with a [Mamba](https://github.com/state-spaces/mamba) SSM block instead of a standard attention block:
+
+```python
+mamba = Mamba(d_model=512, d_state=64, d_conv=4, expand=2)
+```
+
+It is inserted as a residual contribution:
+
+```python
+if isinstance(block, Mamba):
+    x_out = x + run_mamba(block, x.to(torch.bfloat16)).to(x.dtype)
+```
+
+Placing it at the midpoint of the U-Net-style encoder–decoder stack lets the SSM act as a bottleneck that compresses long-range sequence state before the decoder layers unpack it. The surrounding transformer layers handle local attention patterns as usual.
+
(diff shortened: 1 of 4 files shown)


MatoTeziTanka:
### Community Review — Non-Record 16 MB Track : Hybrid XSA-SSM 

```
ModuleNotFoundError: No module named 'mamba_ssm'
```

---
*Reviewed by [@MatoTeziTanka](https://github.com/MatoTeziTanka) — [The Agora](https://matotezitanka.github.io/parameter-golf). CPU smoke test (CT2038 proteus-engine, 2026-04-11): **IMPORT_FAIL** — ModuleNotFoundError: No module named 'mamba_ssm'. Classification via `classify_prs.py` AST-based classifier; full compliance audit deferred until the import issue is resolved. Auto-drafted from a template and spot-checked before posting.*

Jash-Vora:
Thanks @MatoTeziTanka — confirmed. Fix is to wrap the top-level import:

```python
try:
    from mamba_ssm import Mamba
    _MAMBA_AVAILABLE = True
except ImportError:
    _MAMBA_AVAILABLE = False
    Mamba = None
```

And guard the instantiation inside `GPT.__init__`:

```python
if i == self._ssm_layer_idx:
    if not _MAMBA_AVAILABLE:
        raise RuntimeError("mamba_ssm not installed. Set SSM_MID_LAYER=-1 to disable.")
    mamba = Mamba(d_model=model_dim, d_state=ssm_d_state, d_conv=4, expand=2)
    mamba = mamba.to(torch.bfloat16)
    blocks.append(mamba)
```

No nested f-string issues present. Will push shortly.

PR #1529, opened 2026-04-11
Record: ImprovedParallelResiduals, 1.0758 BPB /   2.7789 nats, -0.0020 BPB / -0.0052 nats vs PR   #1523

msisovic:
# Record: Improved Parallel Residuals

**val_bpb: 1.07578747** (3-seed mean, std 0.0007) | **2.77887078 nats** | **~15.98 MB** | 8xH100 SXM, 600s | Legal TTT

This submission starts from [PR #1523](https://github.com/openai/parameter-golf/pull/1523). Most of the newer submissions moved away from my fuller parallel-residual formulation and settled on a simpler GPT-J-style split-lane decoder. This version keeps the strong parts of that newer baseline and reintroduces the useful parts of my parallel residual implementation.

The key architectural change relative to PR #1523 is in the decoder after the split point. Attention and MLP read from different lanes, but neither sublayer writes back immediately. Instead, both outputs are accumulated into the two lanes together at the end of the block:

```python
next_lane0 = attn_resid * lane0 + attn_post[0] * attn_out + mlp_post[0] * mlp_out
next_lane1 = mlp_resid * lane1 + attn_post[1] * attn_out + mlp_post[1] * mlp_out
```

That keeps the GPT-J-style parallel-in-time update, while restoring the richer learned routing between the attention and MLP lanes. The other important part is that decoder U-Net skips are still written only into `lane0`, which preserves the cheaper and more stable skip path from the newer baseline. Attention reads the mixed `lane0/x0` path, while MLP reads raw `lane1`. Final output uses the mean of the two lanes.

In practice, that is pretty much the only modeling change here versus PR #1523, together with moving `PARALLEL_RESIDUAL_START` from the baseline's `7` to `8`. I ablated that start-layer change separately on top of the plain PR #1523 baseline, without my fuller parallel residual routing changes, and it gave a mild regression on its own. The other notable requirement is that I needed the CUTLASS EVT path to recover the full throughput. In this iteration the CUDA/C++ source is inlined into the training script itself and built against a standard `/opt/cutlass` checkout rather than shipping a separate prebuilt `.so`.

## Results (8xH100 80GB SXM, 600s)

| Seed | Steps | ms/step | Post-EMA BPB | Legal TTT BPB | val_loss (nats) | Artifact |
|------|-------|---------|--------------|----------------|-----------------|----------|
| 1337 | 4,655 | 126.13 | 1.0830 | **1.0751** | 2.7770 | 15,983,095 |
[...]

PR #1354, opened 2026-04-04
Record: varlen+fused mlp+ttt (bpb=1.1093)

samacqua:
# Record: Varlen attention + fused MLP + TTT

**val_loss: 1.8729 | val_bpb: 1.1093** | **~15.9 MB** | 8×H100 SXM, 600s train + ~500s TTT eval

Increased training speed ~3% via variable length attention and a fused MLP kernel, yielding a ~0.002 nat improvements. Re-added an optimized document-based LoRA TTT that yields a ~0.007 nat improvement. Together, these 3 improve performance ~0.009 nats.

Based on a hackathon last weekend with @aldopareja, @sestinj, and @chrishamblin7 :)

Note: it is very hard to tell what sota is because there are (at time of this PR) 905 open PRs, 279 of which are tagged as records, and *many* of these PRs cheat in some way. Afaict, the three improvements in this PR are orthogonal to any recent PR, so once another record is merged, I will add these changes.

## Main changes

Improves upon record [2026-03-25_ValCalib_GPTQ_XSA_BigramHash3072](https://github.com/openai/parameter-golf/blob/main/records/track_10min_16mb/2026-03-25_ValCalib_GPTQ_XSA_BigramHash3072/README.md) with 3 things:

### 1. Variable length attention (~1% faster training, ~0.001 nats)

Replaced dense causal attention with Flash Attention 3's `flash_attn_varlen_func`. During training, documents are packed into flat token buffers with `cu_seqlens` boundaries so attention is computed within documents only — the model never attends across unrelated documents that happen to be adjacent in a batch.

This does two things:
- Removes the need for the model to learn to ignore pre-BOS content from unrelated documents (val loss is ~.02 nats lower at step 4000).
- Reduces wasted FLOPs: e.g. 10 short (100-token) docs packed into a 1k-token buffer cost proportional to `100 * 100**2 = 1M` attention FLOPs vs `10 * 1000**2 = 10M` with dense attention. This leads to ~1% faster training on 8xH100, and the additional training steps buy ~0.001 nats improvement. This improvement is limited because the model is so small that there is a lot of overhead which is not in the attention so it can only be sped up so much.

### 2. Fused MLP (~1% faster training, ~0.001 nats)

[...]

Issue #1409, opened 2026-04-06
Are HW optimization solutions also welcome?


EOBD:
Hi guys!

thank you, first, for hosting this comp🙏 a really cool idea

Maybe it's just me, but using about half a tera of HBM for a 30MB of weights just doesnt sit right.... (:
And I though of trying to stuff everything in L1/SH, and pipelining the model across weights across the 8 pieces. Than "PP" the forward and backwards passes. passing the activations through HBM. (a persistant kernel with HBM/L2 synchronization. could really nicely hide HBM trips. and updating the weights in SRAM directly is also super cool)
Is this a direction you'd want to see? (where it's more about the basics and exploring new utilization paradigms – than cutting edge arch theory)

THANKS!

Issue #1017, opened 2026-03-28
A Field Guide to Valid Submissions


NoesisGenesis:
> This is a first draft. I am not happy with all of it, but my weekend has other claims on it, so revisions will come when they come. I would rather this exist imperfectly now than perfectly later. Most but not all of the writing below is my own. Where the prose quality dips, blame the AI, not me. Headings and formatting were mostly delegated. The parts that matter most are entirely mine. I expect a significant share of this document's value to come from agents reading it and, ideally, stopping themselves from producing the same invalid submissions they have been producing at scale.

## How We Got Here

The parameter-golf repository has a PR problem, and if you have been paying attention, you already know what kind. 

When you combine a public leaderboard with a README that mentions hiring and the tantalizing possibility that someone at OpenAI might glance at your GitHub profile, you get an incentive gradient steep enough to produce exactly the kind of behavior that has, in fact, been produced.   

The chain of events is not complicated:   

1. People want to be noticed. Some of them are, if we are being honest, less focused on the compression metric than on compressing the distance between themselves and an OpenAI offer letter one impressive PR at a time. This is not an unreasonable hope. It is, however, an unreasonable strategy when the PRs you submitted as a demonstration of competence are, in fact, a demonstration of the opposite.

2. To claim a record, you must beat the current rank 1 by at least 0.005 nats with statistical significance. PRs are reviewed chronologically. So if your improvement is small, you are in a race you cannot see against competitors you cannot count, any one of whom might, at any moment, render your contribution permanently invisible. The PR feed is a constant reminder that someone, somewhere, might be about to do exactly this. 

3. The natural response to this pressure is to delegate to something faster than yourself, which in practice means an AI agent, a harness of agents, or some other configuration of automated ambition pointed at the repository. The natural workflow is to let it read the top PRs, synthesize the best ideas, and produce a submission. The natural outcome is a PR that reports a suspiciously good number, and contains at least one error that requires domain knowledge to detect and neither the agent nor its operator noticed. The operator, _of course_, would have caught it themselves. They simply had the misfortune of not looking before they submitted. In their defense, looking takes time, and time was the thing they were trying to save.  

4. The result is what you get when a hundred agents independently read the same top PRs and independently arrive at the same synthesis: a flood of submissions that appear distinct, share the same structural flaws, and reproduce the same evaluation bugs with impressive fidelity. The agents are just not original enough to fail differently.

[...]

PR #1530, opened 2026-04-11
Record: Varlen attention + fused MLP + doc-independent TTT (1.07336)

samacqua:
# Record: Varlen attention + fused MLP + TTT

**val_loss: 2.77261 | val_bpb: 1.07336** | **~15.99 MB** | 8×H100 SXM, 587s train + ~340s TTT eval
| Seed | BPB | Loss |
|------|-----|------|
| 0 | 1.07258208 | 2.77059090 |
| 1 | 1.07324696 | 2.77230836 |
| 2 | 1.07426259 | 2.77493185 |
| **Mean** | **1.07336388** | **2.77261037** |
| **Std** | **0.00084633** | **0.00218618** |

Best PR bpb ([PR #1529](https://github.com/openai/parameter-golf/pull/1529)): bpb=1.0753 (**delta=0.0019**), loss=2.7776 (**delta=0.0050**)

Merged record bpb ([PR #1493](https://github.com/openai/parameter-golf/pull/1493)): bpb=1.0810 (**delta=0.0076**), loss=2.7923 (**delta=0.0197**)

Increased training speed ~5% via variable length attention, a fused MLP triton kernel (no `cutlass_evt_fusion` dep), and grouping together small parameters, yielding ~.002 nats when comparing sliding window eval. Re-added document-based LoRA TTT which has *no inter-sequence dependence* and improves over strided evaluation by ~.008 nats.

Based on a hackathon last weekend with @aldopareja, @sestinj, and @chrishamblin7 :)

## Main changes

Applied changes from [my old PR](https://github.com/openai/parameter-golf/pull/1354) to a recent record PR: [#1523](https://github.com/openai/parameter-golf/pull/1523). But [PR #1552](https://github.com/openai/parameter-golf/pull/1552) beat my previous bpb before I submitted the PR, so I incorporated their (orthogonal) improvements. Most of below is copied from my previous PR [#1354](https://github.com/openai/parameter-golf/pull/1354).

This involves 3 things:

### 1. Variable length attention (~2% faster training, ~0.001 nats)

Replaced dense causal attention with Flash Attention 3's `flash_attn_varlen_func`. During training, documents are packed into flat token buffers with `cu_seqlens` boundaries so attention is computed within documents only — the model never attends across unrelated documents that happen to be adjacent in a batch.

This does two things:
- Removes the need for the model to learn to ignore pre-BOS content from unrelated documents
- Reduces wasted FLOPs: e.g. 10 short (100-token) docs packed into a 1k-token buffer cost proportional to `100 * 100**2 = 1M` attention FLOPs vs `10 * 1000**2 = 10M` with dense attention.

### 2. Fused MLP + grouped small params (~3% faster training, ~0.001 nats)

A custom Triton kernel (`linear_leaky_relu_square_kernel`) fuses the up-projection, LeakyReLU(0.5)² activation, and squaring into a single kernel. Based on similar kernels from [modded-nanogpt](https://github.com/KellerJordan/modded-nanogpt/blob/master/triton_kernels.py). I also group the many tiny replicated scalar/control gradients into a single all-reduce to avoid a pile of tiny collectives.

### 3. Doc-based test-time training (TTT) (~0.008 nats over sliding window)

> [Blog explaining LoRA-based TTT from past record](https://samacquaviva.com/projects/parameter-golf/)

Although it is technically legal in this competition to train on tokens from previous documents in the dataset, I am spiritually opposed to this. Under the current formulation, if the eval set was bigger, the expectation of the loss would be lower which seems broken. So in this implementation, there is score-first TTT applied to each sequence in the validation set *independently* (and efficiently using batched LoRAs), which is strictly harder.

Re-adds LoRA-based TTT, based on [my old implementation](https://github.com/openai/parameter-golf/blob/main/records/track_10min_16mb/2026-03-17_LoRA_TTT/README.md), but > 2x faster which allows for using smaller chunk sizes which leads to better performance. This is an instance of "Case 3" according to [this classification](https://samacquaviva.com/projects/ttt-clarification/).

It's interesting to note that adding test-time training improves loss more than adding ~215 steps. These 215 steps train on `786432*215=169,082,880` tokens to gain ~.002 nats. The average sequence length in the validation set is ~200 tokens which means test-time training here gains ~.003 nats / 800 tokens on average (valid bc sequences are trained independently). So, in a way, TTT is `~(.003/800) / (.002/169082880) >= 300k` times more token efficient than pre-training: it helps to be in distribution :)

## Other small changes

Made some changes to make replication and dev based on this PR easier:

- Load from a checkpoint just for eval
- Didn't submit minified code, instead wrote that utility into the script when calculating file size so that it is easier for people to build off of this
- Store unminified code in logs

## Replicating runs + dev

```bash
# setup
uv venv
source .venv/bin/activate
uv pip install -r records/track_10min_16mb/2026-04-10_VarLenAttn/requirements.txt
uv pip install --break-system-packages flash_attn_3 --find-links https://windreamer.github.io/flash-attention3-wheels/cu128_torch291
uv pip install torch==2.9.1+cu128 --extra-index-url https://download.pytorch.org/whl/cu128

  python3 data/cached_challenge_fineweb.py --variant sp8192 --train-shards  128

# train + eval
SEED=0
ARTIFACT_DIR="runs/varlen${SEED}" SEED=$SEED \
    torchrun --standalone --nproc_per_node=8 \
    records/track_10min_16mb/2026-04-10_VarLenAttn/train_gpt.py

# eval saved checkpoint w/ TTT (useful for dev)
EVAL_ONLY_PATH="runs/varlen${SEED}/final_model.pt" SEED=$SEED \
    torchrun --standalone --nproc_per_node=8 \
    records/track_10min_16mb/2026-04-10_VarLenAttn/train_gpt.py
```

### DIFF
--- a/records/track_10min_16mb/2026-04-10_VarLenAttn/README.md
+++ b/records/track_10min_16mb/2026-04-10_VarLenAttn/README.md
@@ -1,0 +1,72 @@
+# Record: Varlen attention + fused MLP + TTT
+
+**val_loss: 2.7806 | val_bpb: 1.07643** | **~15.99 MB** | 8×H100 SXM, 600s train + ~360s TTT eval
+| Seed | SW Loss | SW BPB | TTT Loss | TTT BPB |
+|------|---------|--------|----------|---------|
+| 0 | 2.78822313 | 1.07940806 | 2.78138792 | 1.07676194 |
+| 1 | 2.78698310 | 1.07892801 | 2.78033428 | 1.07635404 |
+| 2 | 2.78652675 | 1.07875134 | 2.77993034 | 1.07619767 |
+| **Mean** | **2.78724** | **1.07903** | **2.78055** | **1.07644** |
+
+Best PR bpb ([PR #1523](https://github.com/openai/parameter-golf/pull/1523)): 1.0778. **delta=.0014**
+Merged record bpb ([PR #1493](https://github.com/openai/parameter-golf/pull/1493)): 1.0810. **delta=.0047**
+
+Increased training speed ~5% via variable length attention, a fused kernel, and grouping together small parameters, yielding ~.002 nats when comparing sliding window eval. Re-added document-based LoRA TTT which has *no inter-sequence dependence* and beats previous record by ~.003 nats.
+
+## Main changes
+
+Applied changes from [my old PR](https://github.com/openai/parameter-golf/pull/1354) to a recent record PR: [Record: SP8192 + Triple Recurrence + Banking + Fused MLP + Muon 0.97 — val_bpb 1.0778 (3-seed mean) #1523](https://github.com/openai/parameter-golf/pull/1523). Most of below is copied from my previous PR.
+
+This involves 3 things:
+
+### 1. Variable length attention (~2% faster training, ~0.001 nats)
+
+Replaced dense causal attention with Flash Attention 3's `flash_attn_varlen_func`. During training, documents are packed into flat token buffers with `cu_seqlens` boundaries so attention is computed within documents only — the model never attends across unrelated documents that happen to be adjacent in a batch.
+
+This does two things:
+- Removes the need for the model to learn to ignore pre-BOS content from unrelated documents
+- Reduces wasted FLOPs: e.g. 10 short (100-token) docs packed into a 1k-token buffer cost proportional to `100 * 100**2 = 1M` attention FLOPs vs `10 * 1000**2 = 10M` with dense attention.
+
+### 2. Fused MLP + grouped small params (~3% faster training, ~0.001 nats)
+
+A custom Triton kernel (`linear_leaky_relu_square_kernel`) fuses the up-projection, LeakyReLU(0.5)² activation, and squaring into a single kernel. Based on similar kernels from [modded-nanogpt](https://github.com/KellerJordan/modded-nanogpt/blob/master/triton_kernels.py). I also group the many tiny replicated scalar/control gradients into a single all-reduce to avoid a pile of tiny collectives.
+
+### 3. Doc-based test-time training (TTT) (~0.003 nats)
+
+> [Blog explaining LoRA-based TTT from past record](https://samacquaviva.com/projects/parameter-golf/)
+
+Although it is technically legal in this competition to train on tokens from previous documents in the dataset, I am spiritually opposed to this. Under the current formulation, if the eval set was bigger, the expectation of the loss would be lower which seems broken. So in this implementation, there is score-first TTT applied to each sequence in the validation set *independently* (and efficiently using batched LoRAs), which is strictly harder.
+
+Re-adds LoRA-based TTT, based on [my old implementation](https://github.com/openai/parameter-golf/blob/main/records/track_10min_16mb/2026-03-17_LoRA_TTT/README.md), but > 2x faster which allows for using smaller chunk sizes which leads to better performance. This is an instance of "Case 3" according to [this classification](https://samacquaviva.com/projects/ttt-clarification/). The TTT gain is only ~.003 nats over sliding window evaluation, as compared to ~.007 in [my previous PR](https://github.com/openai/parameter-golf/pull/1354) and ~.006 in the most recent record which trains on the entire validation sequence. This gap is probably closeable with better hparams, I largely just took my hparams optimized for the non-looped transformer.
+
+It's interesting to note that adding test-time training improves loss more than adding ~215 steps. These 215 steps train on `786432*215=169,082,880` tokens to gain ~.002 nats. The average sequence length in the validation set is ~200 tokens which means test-time training here gains ~.003 nats / 800 tokens on average (valid bc sequences are trained independently). So, in a way, TTT is `~(.003/800) / (.002/169082880) >= 300k` times more token efficient than pre-training: it helps to be in distribution :)
+
+## Other small changes
+
+- Added some useful dev things, like loading from a checkpoint just for eval
+- Didn't submit minified code, instead wrote that utility into the script so that it is easier for people to build off of this
+
+## Replicating runs + dev
+
+```bash
+# setup
+uv venv
+source .venv/bin/activate
+uv pip install -r records/track_10min_16mb/2026-04-10_VarLenAttn/requirements.txt
+uv pip install --break-system-packages flash_attn_3 --find-links https://windreamer.github.io/flash-attention3-wheels/cu128_torch291
+uv pip install torch==2.9.1+cu128 --extra-index-url https://download.pytorch.org/whl/cu128
+
+MATCHED_FINEWEB_REPO_ID=kevclark/parameter-golf \
+  python3 data/cached_challenge_fineweb.py --variant sp8192 --train-shards  128
+
+# train + eval
+SEED=0
+ARTIFACT_DIR="runs/varlen${SEED}" SEED=$SEED \
+    torchrun --standalone --nproc_per_node=8 \
+    records/track_10min_16mb/2026-04-10_VarLenAttn/train_gpt.py
+
+# eval saved checkpoint w/ TTT (useful for dev)
+EVAL_ONLY_PATH="runs/varlen${SEED}/final_model.pt" SEED=$SEED \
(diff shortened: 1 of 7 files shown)


dexhunter:
I may be missing something, but I think there is one higher-scrutiny `#1017` / README issue worth clarifying.

In the TTT path, the compile warmup appears to use actual validation tokens before the main eval loop, and it also does `backward()` / `step()` inside that warmup block. The main score loop itself looks score-first, so this is not a claim about the core TTT logic; the concern is specifically the pre-eval warmup.

My read of the current guidance is:
- Track B allows score-first adaptation using previously scored eval tokens
- but not adaptation on validation tokens before they are scored

If that reading is right, would you be willing to switch the warmup to:
- synthetic tokens / shape-only warmup, or
- training tokens, or
- a no-update warmup

That would make the legality story much cleaner.

samacqua:
@dexhunter it could honestly just be commented out, given that warmup + eval time is still < 600s. But it shouldn't matter -- training warmup does the same thing, parameters and optimizer states are reset. As a sanity check I re-ran TTT on seed 2 w/ warmup commented out, and the loss was within expected variance between runs (actually did slightly better): `quantized_ttt_lora val_loss:2.77492177 val_bpb:1.07425869 eval_time:338465ms`.

But given that making a change + re-running what take an hour of 8xh100, I will only if it is a blocker.

MatoTeziTanka:
### Community Review — VarLen attention + fused MLP + doc-independent TTT

Thanks @samacqua. Doc-independent TTT via cu_seqlens boundary isolation is a genuinely interesting approach to the causal-dependence question the SLOT cluster has been bouncing around. One import blocker, then a deeper question on the doc-independence claim.

**What I found** (head SHA `161d64428159c61f2d42dd6d415ec1386599ef90`, `records/track_10min_16mb/2026-04-10_VarLenAttn/train_gpt.py`, 116,694 bytes of actual source — not shim-compressed, directly readable):

- **Imports (L1-12):** `from flash_attn_interface import (flash_attn_varlen_func, ...)` at the top — hard import, no fallback
- `eval_val_sliding` at L1948 — uses `BOS_ID` to find document boundaries via `(chunk_cpu[:-1] == BOS_ID).nonzero()`, builds `cu_seqlens` from those boundaries, then calls the FA3 varlen kernel with that cu_seqlens argument. This is standard document-packed varlen attention — within a batch, attention can't cross a BOS boundary.
- `_build_cu_seqlens(bos_pos, total_len, device, max_doc_len=0, bucket_size=64)` at L260 — helper that builds the `cu_seqlens` tensor with a bucket-size 64 padding
- `DocumentPackingLoader` at L283 — the training-time document-packed loader
- Standard `GPT` at L730 with varlen-aware blocks

**"Doc-independent TTT" — the interesting idea.** My read is that if the LoRA (or whatever TTT-like adaptation you're running) respects the same `cu_seqlens` boundaries as attention, then when token `t` in document D is scored, the adaptation state derived from document D' ≠ D doesn't influence `t`'s scoring through the attention path — because attention physically can't cross the boundary. That's a clean causal isolation argument IF the adaptation state also respects document boundaries.

The open question is whether the adaptation state itself is per-document or per-batch. I couldn't find an `eval_val_sliding_ttt` / `eval_val_ttt` / `eval_val_slot` function in the 116KB source via my structural grep — could you point me at where the TTT adaptation lives in this codebase? "doc-independent TTT" in the title suggests there's an eval-time adaptation somewhere, but the function I was expecting by name doesn't show up. Is the TTT-like adaptation integrated into the main eval path, or does it use a different function name?

**Import blocker (smoke test).** The CPU smoke on CT2038 hit:
```
IMPORT_FAIL error=ImportError("cannot import name 'flash_attn_varlen_func' from 'flash_attn_interface' (unknown location)")
```
My flash_attn stub covers `flash_attn_func` but not the varlen variant. This is a CPU-stub limitation, not a PR defect — FA3 is available on H100s where this is intended to run.

### Questions

1. **Where is the doc-independent TTT adaptation loop?** `eval_val_sliding` at L1948 looks like standard no-grad scoring. Is the adaptation inlined into the forward pass, or does it live in a separately-named function I missed?
2. **Per-batch vs per-doc adaptation state**: if documents D_i and D_j are packed into the same batch with disjoint cu_seqlens, does the TTT state they each produce stay isolated, or does it mix before being used to score the next batch?
3. **Cross-document information flow through DocumentPackingLoader**: the training-time loader packs multiple documents. Does the TTT adapt at training time (which would be legal as a training-side technique) or at eval time (which needs the doc-independence argument to close)?

### Compliance summary (partial)

- **N-gram family bug:** not present (no `full_key` / `ctx_hash ^ target * primes`)
- **Scored-region SLOT:** not present (no `slot_loss` / mask = scored region)
- **Pre-Quant TTT on val_tokens:** not present (no `prequant_ttt_adapt_adamw`)
- **Varlen attention via FA3 cu_seqlens:** legitimate hardware optimization per Issue #1409
- **Doc-independent TTT causal argument:** pending clarification on the adaptation loop location

**Verdict:** LOOKS INTERESTING, NEEDS AUTHOR CLARIFICATION on the TTT adaptation path.

**Recommendation to @cocohearts @valerio-oai @0hq @yuzhougu-oai @notapplica:** HOLD pending author clarification on where and how the doc-independent TTT runs. If the adaptation respects cu_seqlens boundaries and the temporal ordering is score-before-adapt at the document level, this is a genuinely clean path out of the SLOT compliance bind, and I'd flip to MERGE.

---
*Reviewed by [@MatoTeziTanka](https://github.com/MatoTeziTanka) — [The Agora](https://matotezitanka.github.io/parameter-golf). CPU smoke test (CT2038 proteus-engine, 2026-04-11): IMPORT_FAIL due to `flash_attn_varlen_func` missing from my flash_attn stub (known stub limitation, not a PR defect). Static review of the 116,694-byte source performed against the compliance axes above — 4 of 5 clean on the standard audits, 1 needs author clarification on the TTT adaptation location. AI tooling: review drafted with Claude Code (Opus); batch-9 subagent quota exhausted mid-batch so this review was authored in the main session with reduced audit depth — the flagged questions above are places I'd want confirmation from the author rather than statements I'm making from full verification. SHA `161d64428159c61f2d42dd6d415ec1386599ef90`.*

samacqua:
@MatoTeziTanka look at `eval_val_ttt_lora` [here](https://github.com/samacqua/parameter-golf/blob/161d64428159c61f2d42dd6d415ec1386599ef90/records/track_10min_16mb/2026-04-10_VarLenAttn/train_gpt.py#L2156):
- each sequence (split by BOS) has it's own LoRA, so there is no dependence between sequences
- each sequence is split into 32-token chunks. We iterate through the chunks in order, evaluating on chunk `i`, *then* training on chunk `i` before moving on to chunk `i+1`.

So yes, it respects the same document boundaries. It is strictly harder (and more valid imo) than TTT on the full validation sequence autoregressively.

See the "Methods" section of [this blog](https://samacquaviva.com/projects/parameter-golf/) for clarity.

dexhunter:
The pattern structurally matches what @valerio-oai flagged as invalid in #677 ("adapt on validation before the reported eval pass"). Even though LoRA resets per batch, the compile warmup still runs backward+step on val tokens before the eval loop. Since you confirmed the fix is within variance, it would be worth switching to random/synthetic tokens to avoid any ambiguity during review.

msisovic:
> The pattern structurally matches what @valerio-oai flagged as invalid in #677 ("adapt on validation before the reported eval pass"). Even though LoRA resets per batch, the compile warmup still runs backward+step on val tokens before the eval loop. Since you confirmed the fix is within variance, it would be worth switching to random/synthetic tokens to avoid any ambiguity during review.

This submission actually looks good to me. They don't "adapt on validation before the reported eval pass", as the warmup/compilation throws away the updates. The final result wouldn't change at all if they replaced those validation tokens in warmup with any other tokens. The author even notes that the result is unchanged, even when they comment out the warmup.

msisovic:
Hey btw @samacqua the training script will crash without the `pyminify` CLI tool installed on the machine, so you might want to add a step to the README that it should be installed (maybe I have missed it though).

samacqua:
@msisovic fixed. Thanks!

PR #1618, opened 2026-04-14
ConvCompressed Residual Language Model

Magumbian:
Simple convolutional compression scheme


Magumbian:
The LLM output is obtained by merging the outputs of smaller mini transformer blocks. In this case the smaller transformer blocks are obtained by convolutional compression of the embedding layer before attention to reduce dimensionality of attention layers, the outputs are then combined with fused by learned softmax scalar weights (NodeFusion). In addition I use depth recurrence to increase the effective depth of the network

PR #1545, opened 2026-04-11
Record: GDN-Hybrid (Gated DeltaNet + SWA) — val_bpb 1.028308 (3-seed cold-cache mean)

Abhishek8108:
## Summary

**val_bpb = 1.028308** (3-seed cold-cache mean) | 14.48–14.70 MB | 8×H100 SXM

First non-transformer architecture in the 10-min record track. Beats merged SOTA (PR #1493, 1.0810) by **5.27 centiBPB**. No TTT. Fixed predictor.

## Architecture

GDN-Hybrid replaces the transformer backbone with **Gated DeltaNet** (delta-rule linear recurrence) + **Sliding Window Attention**:

```
[GDN×5] → [SWA] → [GDN×5] → [SWA_shared]
```

- 33,862,953 params, 512-dim, SP1024 tokenizer
- GDN: `fla.layers.GatedDeltaNet`, head_dim=64, use_short_conv=True
- SWA: window=512, 8 heads / 4 KV heads, weight-shared across both layers
- QK-Gain 5.0, BigramHash(3072×112) + trigram embeddings, logit softcap 30.0
- Quantization: full-Hessian int6 GPTQ + zstd-22

## 3-Seed Results (cold-cache, fresh pods)

| Seed | Steps | EMA BPB  | **Quantized BPB** | Artifact (bytes) |
|------|-------|----------|-------------------|-----------------|
| 42   | 1857  | 1.017970 | **1.027163**      | 15,188,240      |
| 1337 | 1858  | 1.018624 | **1.027614**      | 15,417,768      |
| 2024 | 1858  | 1.020559 | **1.030148**      | 15,314,099      |
| **Mean** | — | **1.019051** | **1.028308** | — |
| **Std**  | — | 0.001356 | 0.001610 | — |

All three seeds run on separate fresh pods. Cold-start signature confirmed (step 1 at ~105s, Triton JIT overhead). t-stat = 56.7 vs SOTA threshold, p ≪ 0.01.

## Reproduction

```bash
pip install flash-linear-attention zstandard sentencepiece
python3 data/cached_challenge_fineweb.py --variant sp1024

SEED=42 ARCH_MODE=D MAX_WALLCLOCK_SECONDS=590 ITERATIONS=9999 \
  TRAIN_SEQ_LEN=2048 TRAIN_BATCH_TOKENS=786432 \
  QK_GAIN_INIT=5.0 GPTQ_ENABLED=1 VAL_LOSS_EVERY=9999 \
[...]

PR #1576, opened 2026-04-12
Record: GDN-Hybrid + Sliding Window Attention + compressed-code warmdown1000 - val_bpb 1.01671 (3-seed mean)

joshkmartinez:
# Summary

val_bpb = 1.01671233 (3-seed mean, std 0.00134386)
15.71–15.90 MB

Improves the GDN-Hybrid fixed-predictor line with a warmdown1000 schedule and compressed-code packaging w/o eval-time adaptation.

| Seed | Steps | EMA BPB | val_bpb | XSA BPB | Artifact bytes |
|------|------:|--------:|--------:|--------:|---------------:|
| 42   | 2227  | 1.007164 | 1.016200 | 1.021202 | 15,733,879 |
| 1337 | 2242  | 1.007164 | 1.015700 | 1.020105 | 15,903,365 |
| 2024 | 2227  | 1.009032 | 1.018237 | 1.024111 | 15,713,422 |
| Mean | —     | 1.007787 | 1.01671233 | 1.021806 | 15,783,555.33 |
| Std  | —     | — | 0.00134386 | — | — |

# Architecture / Technique Stack

1. SP1024 tokenizer
2. GDN-Hybrid backbone: [GDN×5] → SWA → [GDN×5] → SWA_shared
3. Fixed-predictor evaluation path (no TTT / no SLOT / no eval-time adaptation)
4. MuonEq-R + AdamW training mix
5. EMA = 0.997
6. warmdown = 1000
7. GPTQ int6 + zstd-22 packaging
8. Compressed-code packaging for train_gpt.py / architectures.py / configs.py to recover artifact headroom

# Compliance

- Fixed-predictor / Track A style submission
- No TTT
- No SLOT
- No RLS
- No eval-time adaptation
- All three artifacts under 16,000,000 bytes
- Training run stays within the 10-minute 8xH100 submission budget

# Notes

XSA telemetry is reported for completeness, but the submitted score is the fixed-model quantized_bpb result above.

# Credits

- PR #1545 for the GDN-Hybrid architecture line
- PR #1413 for the compressed self-extracting LZMA pattern

PR #1632, opened 2026-04-15
Record: GDN-Hybrid (Gated DeltaNet + SWA) — val_bpb 1.0274 (2-seed mean)

Hkoyuer:
## Summary
- GDN-Hybrid reproduction: **val_bpb 1.0274** (2-seed cold-cache mean)
- 2 seeds (1337, 42) on fresh 8xH100 SXM pods, 590s training each
- All artifacts under 16MB (14.6-14.8 MB)
- No TTT, fixed predictor

## Results

| Seed | Steps | EMA BPB | Quantized BPB | Artifact |
|------|-------|---------|---------------|----------|
| 1337 | 1857 | 1.018060 | **1.026927** | 15,524,240 |
| 42 | 1856 | 1.018499 | **1.027811** | 15,305,698 |
| **Mean** | — | **1.018280** | **1.027369** | — |

## Novel Research
Includes ECN (Error Correction Network) research achieving -0.039 BPB at zero artifact cost. See ECN_RESEARCH.md.

## Test plan
- [x] 2-seed cold-cache training on 8xH100 SXM
- [x] All artifacts under 16MB
- [x] All runs under 590s training
- [x] Round-trip GPTQ quantization verified
- [ ] Judges verify reproducibility

Based on GDN-Hybrid architecture from PR #1545 by @dexhunter.

**Author:** Hamza Koyuer (@Hkoyuer) — [Helolinks.com](https://helolinks.com)

### DIFF
--- a/records/track_10min_16mb/2026-04-15_GDN_Hybrid_DeltaRule_1.028/ECN_RESEARCH.md
+++ b/records/track_10min_16mb/2026-04-15_GDN_Hybrid_DeltaRule_1.028/ECN_RESEARCH.md
@@ -1,0 +1,86 @@
+# Error Correction Network (ECN) — Novel Research
+
+## Concept
+
+A tiny neural network (~1000 parameters) that corrects the logits of the main model during evaluation, learning per-token from its own prediction errors.
+
+The network learns: "when the model is uncertain AND recent accuracy is dropping, shift probability toward frequent tokens." These are non-linear interactions that simple bias correction cannot capture.
+
+## Results
+
+| Method | BPB Improvement | Speed | Usable in 10min? |
+|--------|----------------|-------|-----------------|
+| ECN (backprop) | **-0.039** | 28 wps | No |
+| ECN freeze-after-warmup | -0.006 | 28 wps | No |
+| BCC (binned context correction) | -0.002 | 1107 wps | Borderline |
+| Bias correction | 0.000 | 608 wps | No |
+| Hybrid bigram | +0.001 (worse) | 1200 wps | No |
+| Ridge regression (FEC) | 0.000 | 1075 wps | No |
+| Self-refine (2-pass) | +0.003 (worse) | 147 wps | No |
+
+## Key Finding
+
+The ECN achieves **-0.039 BPB** — larger than most TTT implementations in this challenge — at **zero artifact cost**. The correction network is created in code and learns during evaluation. No weights stored.
+
+The fundamental bottleneck is PyTorch autograd overhead on per-token updates, not the mathematics (which is only ~2000 FLOPs per token for 1000 parameters).
+
+## Why It Works
+
+The ECN does NOT predict the next token (the main model already does that better than any simple corrector). Instead it learns the model's **systematic errors**:
+
+- Underprediction of common tokens after punctuation
+- Overconfidence on rare tokens in low-entropy contexts
+- Calibration drift over long documents
+
+This is fundamentally different from:
+- **Hybrid/ensemble approaches** (which failed — mixing with a weaker predictor always hurts)
+- **Simple bias correction** (which lacks the non-linear feature interactions)
+- **Pre-trained frozen correction** (which can't adapt to the specific validation data)
+
+## Future Work
+
+- Custom Triton/CUDA kernel for fused forward+backward could achieve 100x speedup
+- ELM + Top-K RLS (frozen random features + recursive least squares on top-32 logits) as gradient-free alternative — currently in development, numerically challenging but promising
+- If eval time limit were extended to 30 minutes, ECN would achieve ~1.028 - 0.039 = **0.989 BPB**
+
+## Additional Research: Adapters on Random Linear Maps
+
+First implementation of "Learning adapters on random linear maps" — an item on OpenAI's README wishlist.
+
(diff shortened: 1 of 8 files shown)


dexhunter:
Two things worth flagging:

**Attribution** — the PR credits PR #1545 to me, but that submission is by @Abhishek8108 (looks like an LLM hallucination).

**BPB bug** — this inherits the same leading-space double-counting bug that led to PR #1545 being closed:

```python
if piece.startswith("▁"):
    base_bytes[i] = len(piece[1:].encode("utf-8")) + 1  # space byte baked in
```

The `+1` bakes the space byte in, and `has_leading_space_lut` adds it again in eval. With ~65% of SP1024 tokens carrying `▁`, the byte denominator is inflated ~14%, deflating the reported BPB by the same factor.

@Abhishek8108 acknowledged this on closing PR #1545:
> The corrected BPB is ~1.18, not 1.028.

Independent reproduction with the same GDN-Hybrid architecture lands at ~1.26 BPB across six config variants on SP1024 — consistent with the corrected range, and well above the merged record at 1.0810.

Worth re-scoring with the reference LUT before this is treated as a record.

Hkoyuer:
Thanks @dexhunter for catching this. Rescoring with 
the corrected byte counting. Updated results coming.

tejasnaladala:
Hey @Hkoyuer, I think this PR has the same `build_sentencepiece_luts` byte-counting bug that led to #1545 and #1576 being closed. Flagging it because the claimed `val_bpb 1.0274` comes from the same byte accounting path.

### The bug, in this PR

`records/track_10min_16mb/2026-04-15_GDN_Hybrid_DeltaRule_1.028/train_gpt.py`, lines 207 to 220:

```python
def build_sentencepiece_luts(sp, vocab_size, device):
    base_bytes = torch.zeros(vocab_size, dtype=torch.float32, device=device)
    has_space = torch.zeros(vocab_size, dtype=torch.bool, device=device)
    is_boundary = torch.zeros(vocab_size, dtype=torch.bool, device=device)
    for i in range(vocab_size):
        piece = sp.id_to_piece(i)
        raw = piece.encode("utf-8")
        base_bytes[i] = len(raw)
            has_space[i] = True
            base_bytes[i] = len(piece[1:].encode("utf-8")) + 1   # +1 is the space byte
        if sp.is_control(i) or sp.is_unknown(i):
            is_boundary[i] = True
    return base_bytes, has_space, is_boundary
```

Then the eval loop at lines 376 to 377 adds the same space byte a second time:

```python
tb = base_bytes_lut[tgt].to(torch.float64)
tb += (has_leading_space_lut[tgt] & ~is_boundary_token_lut[prev]).to(torch.float64)
```

So for any `(prev, tgt)` pair where `tgt` starts with `\u2581` and `prev` is not a boundary token, the byte count includes the space byte twice. `byte_count` gets inflated and the reported `BPB = loss / byte_count` comes out lower than the reference metric would give.

### The main-branch reference

`train_gpt.py` on `main` handles this correctly, around lines 180 to 202:

```python
def build_sentencepiece_luts(...):
    ...
    for token_id in range(sp_vocab_size):
        ...
        piece = sp.id_to_piece(token_id)
            has_leading_space_np[token_id] = True
            piece = piece[1:]                     # strip the underscore
        base_bytes_np[token_id] = len(piece.encode("utf-8"))  # no +1
```

The `piece = piece[1:]` strips the underscore, `base_bytes` stores only the token content, and the eval loop adds the space byte exactly once when the previous token is not a boundary. One path one addition.

### Expected magnitude

For SP1024 on FineWeb a typical `\u2581`-prefixed token has maybe 3 to 5 content bytes. Adding a second space byte on most non-boundary transitions inflates the denominator by roughly 14 to 18 percent across the validation set, which translates to a reported BPB that is materially lower than the reference would return. @Abhishek8108 wrote on #1545 that the corrected number on that submission was around 1.18 instead of 1.028. Since #1632 inherits the same LUT logic on the same architecture class and tokenizer I expect the corrected #1632 number to land in a similar range. The exact number has to come from a rescore on the submitted artifacts not from me, but the direction and rough magnitude should match.

### Suggested fix

Drop the current `build_sentencepiece_luts` and copy the one from `main` verbatim. Main's version also handles `sp.is_byte`, `sp.is_control`, `sp.is_unknown`, `sp.is_unused`, and sizes the table defensively, which the current version does not fully cover. The eval loop in #1632 does not need changes once the LUT is fixed.

Concretely:

1. Strip `\u2581` before assigning `base_bytes[i]`.
2. Remove the `+ 1`.
3. Let the eval loop's `tb += (has_leading_space_lut[tgt] & ~is_boundary_token_lut[prev])` do the single conditional add.

### Request

Could you rescore the submitted `run_seed_1337` and `run_seed_42` artifacts with the `main`-branch `build_sentencepiece_luts`, or patch the LUT in this PR and update the claimed BPB? Flagging here because #1545 and #1576 already went through the same cycle and I don't want another closed PR if a patch is possible.

Not trying to attack the GDN-Hybrid architecture work, just the inherited accounting. A GDN-Hybrid submission with the correct LUT would still be a useful datapoint on where that architecture actually sits.

Related context: #1545 (closed), #1576 (closed), main branch `train_gpt.py`.

PR #1668, opened 2026-04-16
Adding Parallel Residuals

Pravin-dev06:
Instead of sequential residuals, parallel residuals introduced to the Transformer block


Pravin-dev06 (commented):
Review done

PR #1681, opened 2026-04-16
GDN-Hybrid: Fix warmdown/SWA/QAT timing — 1.0208 BPB

OE-GOD:
## Summary

PR #1545's GDN-Hybrid has three training systems (warmdown, SWA, Late QAT) that are configured but **never activate** due to a timing mismatch:

- `ITERATIONS=9999` + `WARMDOWN_ITERS=3000` → warmdown starts at step 6999
- Training only reaches ~2100 steps in 590s wallclock
- **Result: zero warmdown, zero SWA, zero QAT**

### The fix (2 lines)

```python
# Before (PR #1545):
iterations = int(os.environ.get("ITERATIONS", 9999))
warmdown_iters = int(os.environ.get("WARMDOWN_ITERS", 3000))

# After:
iterations = int(os.environ.get("ITERATIONS", 2200))
warmdown_iters = int(os.environ.get("WARMDOWN_ITERS", 400))
```

This activates all three systems:
- **Warmdown** starts at step 1800 (cosine decay lr 1.0 → 0.0)
- **SWA** triggers at step 2100 (lr_mul < 0.2), collects 3 checkpoints
- **Late QAT** triggers at step 2099, reducing quantization degradation

Also fixes a **SWA device mismatch bug** on line 1037 where `swa_avg` tensors on CPU were mixed with `avg_state` tensors on CUDA.

## Results

8×H100, seed=42, SP1024, `runpod/parameter-golf:latest`, 590s wallclock:

| | Without fix (PR #1545 defaults) | With fix |
|--|--------------------------------|----------|
| Training steps | 2119 | 2200 |
| Warmdown | **None** (never triggers) | Steps 1800-2200 |
| SWA checkpoints | 0 | 3 |
| Late QAT | **Never** | Step 2099 |
| EMA BPB | 1.0109 | **1.0064** |
| Quantized BPB | 1.0243 | **1.0208** |
| Artifact | 14.91 MB | 15.03 MB |

**Improvement: -0.0035 BPB** from fixing the training pipeline alone.

## The insight

The model architecture is already good. The waste is in the pipeline around it — three systems designed to improve final model quality were silently disabled by a config mismatch. Same pattern as PR #1510 (ANS compression): optimize the boring parts.

## How to reproduce

```bash
ARCH_MODE=D VOCAB_SIZE=1024 \
DATA_PATH=./data/datasets/fineweb10B_sp1024 \
TOKENIZER_PATH=./data/tokenizers/fineweb_1024_bpe.model \
[...]

PR #1672, opened 2026-04-16
Record: GDN-Hybrid + TMA Megakernel + Brotli-11 — val_bpb 1.01195 (3-seed mean)

andrewbaggio1:
## Summary

**val_bpb = 1.01195** (3-seed mean, std 0.00061) | 8xH100 SXM, 600s | GDN-Hybrid + Fused MLP Megakernel

This is the first GDN-Hybrid submission with a fused Triton megakernel. It combines the GDN recurrent architecture from PR #1576 with a Hopper TMA-accelerated fused relu_sq MLP kernel, brotli-11 model compression (everyone should use this!!), and GPTQ int6 quantization with eigendecomposition-based Hessian repair.

| Seed | Steps | EMA BPB | **Quantized BPB** | Artifact bytes |
|------|-------|---------|-------------------|---------------|
| 42   | 2,212 | 1.000783 | **1.011245** | 15,765,907 |
| 1337 | 2,736 | 1.001806 | **1.012324** | 15,804,631 |
| 2024 | 2,758 | 1.000638 | **1.012281** | 15,802,165 |
| **Mean** | — | **1.001076** | **1.011950** | — |

## Architecture

GDN-Hybrid Model D: `[GDN x5] -> [SWA] -> [GDN x5] -> [SWA_shared]`

- **GatedDeltaNet** recurrent layers from the FLA library (`fla.layers.GatedDeltaNet`), dim=512, 8 heads, head_dim=64, `allow_neg_eigval=True`
- **Sliding Window Attention** with weight-sharing (Griffin/Zamba-style), window=512, QK-Gain 5.0
- **Fused relu_sq MLP megakernel**: Hopper TMA persistent Triton kernel fusing FC projection + ReLU^2 activation + gradient caching in a single kernel launch per block. BLOCK_M=128, BLOCK_N=128, BLOCK_K=64
- SP1024 tokenizer, 33.5M parameters, bigram path disabled (dead weights)
- SmearGate, logit softcap 30.0, tied embeddings

## Training

- MuonEq-R optimizer (momentum=0.97) + AdamW for scalars/embeddings
- EMA decay=0.997 + SWA (17-18 checkpoint average during warmdown)
- Warmdown schedule: `TOTAL_ITERATIONS=2100`, `WARMDOWN_ITERS=1000` (cosine from lr=1.0 to 0.0)
- Wallclock budget: 590s on 8xH100 SXM

## Quantization & Compression

- **GPTQ int6** with autoregressive self-generated calibration data (64 sequences)
[...]

PR #1687, opened 2026-04-17
Record: K_KVShare_Wider full-recipe FLA — val_bpb 1.04090 (3-seed mean)

resouer:
## Summary

**3-seed mean val_bpb: 1.04089763** (std 0.00106003) | **3.16476760 nats** | **8xH100 SXM, 600s**

| Seed | Quantized BPB | val_loss (nats) | Artifact |
|------|---------------:|----------------:|---------:|
| 1337 | **1.03967403** | 3.16104735 | 15,762,406 |
| 42 | **1.04153708** | 3.16671180 | 15,870,797 |
| 2025 | **1.04148177** | 3.16654364 | 15,648,800 |
| **Mean** | **1.04089763** | **3.16476760** | **15,760,668** |

## Mechanism

This candidate packages the stronger `K_KVShare_Wider` point on a fuller upstream-style FLA / GatedDeltaNet recipe.

Main idea:
- FLA / GatedDeltaNet family
- `K_KVShare_Wider`: KV sharing (`kv_sharing_stride=2`) used to buy width rather than depth
- fuller upstream-style recipe
- EMA + SWA + late QAT + int6 artifact path

Nearest prior family reference: PR `#1370`.

## Compliance notes

What this candidate does **not** use:
- no TTT
- no SLOT
- no n-gram overlay
- no SWA/XSA final scoring path (`K_KVShare_Wider` has `num_swa_layers = 0`)

Hardening already applied:
- all 3 seeds are on one script revision
- `train_gpt.py` does not perform runtime dependency downloads
- dependencies are declared in `requirements.txt` and expected to be installed before evaluation

## Packaging note

This PR keeps the faithful multi-file records surface rather than the tidier single-file experiments.

A direct packaged smoke on this exact multi-file surface (seed `1337`) completed at **1.03971272** BPB, versus the measured `seed1337` result of **1.03967403**, a delta of only **+0.00003869 BPB**. Peak memory remained **41127 MiB**.

## Packaged-folder verification

Exact draft records-folder audit:
- packaged folder + largest artifact = **15,991,282 bytes**
- remaining headroom = **8,718 bytes**

The packaged code in the records folder passes `py_compile` from inside the folder.

## Reproduction

```bash
pip install --no-deps -r requirements.txt
SEED=$SEED ARCH_MODE=K MAX_WALLCLOCK_SECONDS=600 VAL_LOSS_EVERY=0 EVAL_COMPILE_ENABLED=0 \
```

### DIFF
--- a/records/track_10min_16mb/2026-04-16_KKVShareWider_FLA/README.md
+++ b/records/track_10min_16mb/2026-04-16_KKVShareWider_FLA/README.md
@@ -1,0 +1,56 @@
+# Record: K_KVShare_Wider full-recipe FLA
+
+**val_bpb: 1.0409** (3-seed mean, std 0.0011) | **3.1648 nats** | **8xH100 SXM, 600s** | **No TTT**
+
+FLA / GatedDeltaNet candidate using `K_KVShare_Wider` on a fuller
+upstream-style recipe. Nearest prior family reference: PR `#1370`.
+The packaged script avoids runtime dependency downloads from `train_gpt.py`.
+
+## Results
+
+| Seed | Steps | Post-EMA BPB | **Quantized BPB** | val_loss (nats) | Artifact |
+|------|------:|-------------:|------------------:|----------------:|---------:|
+| 1337 | 1652 | 1.020660 | **1.03967403** | 3.16104735 | 15,762,406 |
+| 42 | 1652 | 1.022042 | **1.04153708** | 3.16671180 | 15,870,797 |
+| 2025 | 1583 | 1.023994 | **1.04148177** | 3.16654364 | 15,648,800 |
+| **Mean** | **1629** | **1.022232** | **1.04089763** | **3.16476760** | **15,760,668** |
+
+## Technique
+
+- FLA / GatedDeltaNet family (`K_KVShare_Wider`)
+- KV sharing is used to buy width rather than depth
+- fuller upstream-style recipe
+- EMA + SWA + late QAT + int6 artifact path
+- final scored line in all logs is `final_int6_roundtrip_exact`
+
+Not used:
+- no TTT
+- no SLOT
+- no n-gram overlay
+- no SWA/XSA final scoring path (`K_KVShare_Wider` has `num_swa_layers = 0`)
+
+## Compliance Notes
+
+- train uses `train_files`; scoring uses `val_files`
+- no eval-time adaptation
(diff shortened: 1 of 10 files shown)


bigbag:
Hi @resouer — I think this PR hits the same `build_sentencepiece_luts` byte-counting bug that closed #1545 / #1576 / #1632 / #1672 / #1681. Flagging for visibility.

**Where:** `train_gdn_7k.py` L192–205 sets `base_bytes[i] = len(piece[1:].encode("utf-8")) + 1` for ▁-prefixed pieces — so the leading-space byte is already baked in. Then the scorer at L361–363 adds **another** `+1` via `has_leading_space_lut[tgt] & ~is_boundary_token_lut[prev]`. Every ▁-token ends up counted with one extra byte.

**Reference** (base `train_gpt.py` L180–204) strips the ▁ from `piece` **before** calling `.encode("utf-8")`, so `base_bytes` counts only the content and the scorer's `+1` is the only leading-space accounting.

**Empirical check** on `fineweb_1024_bpe.model` with a 3,648-byte FineWeb-style sample (reproduces the scorer arithmetic from L361–363):

| | actual text | reference LUT | this PR's LUT |
|---|---:|---:|---:|
| byte_count | 3,648 | 3,647 | **4,214  (+15.55%)** |

Per-token delta (PR − ref): 68.0% of tokens → 0, 31.5% → +1 (the ▁-tokens), 0.5% → +5 (looks like `sp.is_byte` tokens, where the reference uses `base_bytes = 1` but this impl uses `len("<0x..>".encode())`).

Correction factor **×1.1555** → the reported 1.0409 3-seed mean becomes **≈ 1.20** under the reference scorer, which is in line with the prior GDN-Hybrid runs before rescoring.

Minimal fix, mirroring base `train_gpt.py` L180–204:
- strip `▁` from `piece` **before** `piece.encode("utf-8")`
- default `is_boundary_token` to `True`; set `False` only for non-control/unknown/unused tokens
- handle `sp.is_byte(i)` → `base_bytes = 1`
- add `sp.is_unused(i)` to the control/unknown skip

Happy to share the verification script if useful.

resouer:
Closing this submission after local verification found a scoring bug in the SentencePiece byte accounting path. The corrected rerun on commit `3bf1ec3` lands at `final_int6_roundtrip_exact val_bpb:1.22285488`, so the reported `1.04089763` result does not hold under the base scorer semantics.

Issue #1604, opened 2026-04-13
Clarify which text normalizations are allowed for custom tokenizers


mikeapedia:
## Context

PR #1578 (Custom Casefold Tokenizer) applies `.lower()` after NFKC normalization before tokenization, which has prompted a productive debate about what normalizations are permitted. The discussion is happening in the PR comments, but the question is broader than any single submission and deserves its own policy thread.

## The question

Custom tokenizers are explicitly allowed (Issues #43, #897). The rules state: *"Instead of locking the tokenizer, we let you bring your own and calculate our validation metrics on the average compression of the validation set."*

But there's no guidance on what text normalization is permitted before or during tokenization. Every SentencePiece submission currently applies `nmt_nfkc` normalization, which was inherited from SP defaults rather than an explicit competition policy.

## Why this matters

Any normalization that collapses distinct representations reduces prediction entropy and makes the modeling task easier. This is a spectrum:

| Normalization | What it collapses | Info loss |
|---|---|---|
| **NFC** (canonical only) | Encoding variants of same character (e + ◌̈ → ë) | None — lossless |
| **NFKC** (current default) | Ligatures, fractions, superscripts, fullwidth chars, math symbols + whitespace collapsing + newline conversion | Lossy |
| **NFKC + case folding** (PR #1578) | Above + uppercase → lowercase | More lossy |
| **Degenerate** (e.g., all text → "a") | Everything | Total |

The competition currently uses NFKC — a lossy normalization — inherited from SentencePiece defaults. Reasonable people can disagree on where to draw the line. The current debate in PR #1578 has surfaced good arguments on both sides.

## Proposal

Could maintainers clarify:

1. **Is NFKC the intended normalization standard, or just an inherited default?** If submissions used NFC instead, would that be allowed? What about `identity` (no normalization)?

2. **Are additional normalizations beyond NFKC permitted?** Case folding, accent stripping, and other lossy normalizations all sit on the same spectrum as NFKC.

3. **Should the policy be documented?** A one-line addition to the README or rules would save future submitters (and reviewers) from re-litigating this for each new tokenizer PR.

Whatever the decision, explicit documentation would benefit the competition. I'll respect whatever the maintainers decide.

SPThole:
In my view, imagine in inference setting, wouldn’t this reduce the effectiveness of a generative model—particularly with transformations like uppercasing → lowercasing?

[...]

PR #1779, opened 2026-04-22
Record: SP8192 + CaseOps + Gated Attention + Quant Gate + Loop4-5 + Phased TTT + Frozen Recurrent Alpha — val_bpb 1.06421

leon2k2k2k:
## Summary

**val_bpb = 1.06421** (3-seed mean, std 0.00023) | **~15.98 MB** | 8×H100 SXM, 600s train / 600s eval | Phased TTT

- Stacks 2 new techniques on the SP8192 CaseOps base (#1736): **frozen learned recurrent alpha/beta** (cross-layer blend scalars trained to convergence then frozen, `RECUR_ALPHA_ENABLED=1 NUM_LOOPS=2`), and **LoRA-TTT improvements** (warm-start A, alpha=144, WD=1.0) from PR #1767
- **−0.00128 BPB** vs base submission PR #1736 (1.06549)
- CaseOps legality pending issue #1604

## 3-Seed Results

| Seed | Pre-TTT BPB | **TTT BPB** | val_loss (nats/tok) | Artifact | Train | Eval |
|------|------------:|------------:|---------------------|----------|------:|-----:|
| 1    | 1.07704     | **1.06395** | 2.32833             | 15,976,882 | 596.1s | 458.7s |
| 777  | 1.07737     | **1.06429** | 2.32906             | 15,975,842 | 596.1s | 458.9s |
| 2025 | 1.07742     | **1.06438** | 2.32927             | 15,976,882 | 596.1s | 453.4s |
| **Mean** | **1.07728** | **1.06421** | **2.32889** | **15,976,535** | 596.1s | 457.0s |
| **Std**  | 0.00021     | **0.00023** | | | | |

All 3 seeds clear both 600s budgets (train + eval) and the 16,000,000-byte decimal artifact cap.

## Key Techniques

1. **SP8192 CaseOps** — Lossless reversible case normalization (TITLE/ALLCAPS/CAPNEXT/ESC operators). Pending #1604.
2. **GatedAttn + QuantGate** (PR #1736) — Full-dim attention gate with int8 passthrough.
3. **Loop4-5 depth recurrence** (PR #1736) — `NUM_LOOPS=2`, recurrence on layers 3–5.
4. **Frozen Recurrent Alpha/Beta** *(this PR)* — Learnable cross-layer blend scalars (`RECUR_ALPHA_ENABLED=1`) trained to convergence then frozen as constants. Converged values baked into the artifact. L4 self-subtract (α=−0.348) acts as a learned gate; L5 aggregates signal from L3+L4.
5. **LoRA-TTT improvements** (PR #1767) — Warm-start A matrix, alpha=144, WD=1.0. Phased score-first TTT (3 phases, 2000 prefix docs).

## Rule Compliance

- Score-first phased TTT (Condition 3), no pre-quant TTT, no n-gram cache
- Artifact ≤ 16 MB (max 15,976,882 B), train ≤ 600s, eval ≤ 600s
- Frozen recurrent scalars are trained weights serialized into the artifact — no parameters outside the 16 MB budget
- CaseOps legality pending issue #1604

## Test Plan

- [ ] Reviewer reproduces any single seed with the provided `train_gpt.py` and env vars from README
- [ ] Verify artifact size `< 16,000,000` bytes in each seed log
- [ ] Verify score-first TTT ordering in code


dexhunter:
Hi @leon2k2k2k — nice work on the frozen recurrent alpha/beta mechanism, interesting pattern. Heads-up that we just patched the same bug in PR #1736 and PR #1769 and wanted to flag it here since this submission forks the same prep script.

**Issue.** `prepare_caseops_data.py` line 157 uses `sp.encode(transformed, out_type=int)` without prepending `BOS_ID = 1`. The SP tokenizer reserves IDs 0–7 (`<pad>`, `<s>`, `</s>`, `<unk>`, TITLE, ALLCAPS, CAPNEXT, ESC), so `sp.encode` can't emit ID 1 naturally. `train_gpt.py:_find_docs` (line 2209) then returns `[]` and `_loss_bpb_from_sums` (line 2303) divides by zero in the phased TTT eval path. Training survives via the `_init_shard:408–409` fallback (`if self.bos_idx.size == 0: self.bos_idx = np.array([0], ...)`); phased TTT has no equivalent fallback.

**Scope.** Prep-script only — the submitted 1.06421 is on valid data if your seed runs used shards produced by an internal pipeline that already prepends BOS. `val_bpb` reduces to `loss_sum/ln(2)/byte_sum` (token counts cancel at line 2303), and `byte_sum` is unchanged with BOS prepended (BOS contributes 0 original bytes). So the metric is fine; only external reproductions crash.

**Fix** — 4-line diff, matches the canonical pattern in `data/download_hf_docs_and_tokenize.py:364–366`:

```python
# near module top, with other constants
BOS_ID = 1

# inside the per-doc loop
for text in _iter_docs(args.docs):
    transformed = encode_lossless_caps_v2(text)
    token_ids = [BOS_ID] + sp.encode(transformed, out_type=int)
    if n_docs < args.val_docs:
        byte_counts = _token_original_byte_counts(sp, text, transformed)
        val_buf_tokens.extend(token_ids)
        val_buf_bytes.append(0)  # BOS = 0 original bytes
        val_buf_bytes.extend(int(b) for b in byte_counts)
    else:
        train_buf.extend(token_ids)
```

Our fix commits if useful for reference — same diff applies cleanly:
- PR #1736: https://github.com/openai/parameter-golf/pull/1736/commits/d7263a3
- PR #1769: https://github.com/openai/parameter-golf/pull/1769/commits/fe7c309

Originally reported by @codemath3000 on PR #1736. Adding a `bos_count > 0` sanity check to the README's reproduction section also helps catch it early.

nprime06:
Flagging the recur_beta / recur_alpha buffers legality:

```python
  self.register_buffer("recur_beta",
      torch.tensor([1.5973426, 1.8826205, 1.9906198], dtype=torch.float32))
  self.register_buffer("recur_alpha",
      torch.tensor([[ 0.251953125, -0.02099609375, -0.01239013671875],
                    [ 0.06689453125, -0.34765625,    0.0031280517578125],
                    [ 0.138671875,   0.2412109375,   0.0272216796875]],
                   dtype=torch.float32))
```

These 12 floats were "trained to convergence then frozen as constants" — i.e. discovered by offline gradient descent on the training loss and baked into the shipped artifact outside the 10-minute training budget. 

Arguments for legal: all 48 bytes are charged against the 16MB cap, they're registered openly as buffers rather than hidden, and the README explicitly permits offline tuning ("Tuning your Adam hyperparameters across a bunch of runs is fine"). 

Arguments for contested: unlike hyperparameters found by search (which encode "what config generally works for this architecture"), these values are gradient-descent outputs on FineWeb training data — they carry task-specific learned signal that required task-relevant gradient steps outside the budget to discover, and the register_buffer + bake-in-converged-values pattern generalizes to arbitrarily many pre-trained weights if precedent is set at 12 floats. Perhaps violates the spirit of the competition.

Just surfacing for clarification.

leon2k2k2k:
@nprime06 thank you so much for flagging it. yes those parameter were learned through an offline session and I wasn't sure if that is allowed.... To me it feels not so different from tuning discrete parameters but of course I get the argument for the other side. 

I think that the data limit does put a natural bound to this type of practice (like in its extremes you can just train for 10 minutes and then put it back in for the live training). Now that I think about it, this same argument can apply to like learn the n-gram on val set before hand, then just put it in as starting data.... so Idk. This definitely should not be allowed on the val set and I think its debatable if this is a good idea on the training set.. But again I could have "just tried a bunch of parameters by hand and picked the ones". So guideline on this would definitely be nice!

leon2k2k2k:
@dexhunter thank you so much for catching this, and amazing work on #1736 and #1769. I’ve patched it now.

jfc43:
One question:

In `prepare_caseops_data.py`, the val_docs is set as 10_000 by default. However, it seems it should be 50_000 based on https://github.com/openai/parameter-golf/blob/main/data/download_hf_docs_and_tokenize.py#L26.

Could you clarify this?

PR #1858, opened 2026-04-27
SP8192 + Score-First TTT + QK-Gain 5.25 — Neural-Only val_bpb 1.0810 (3-seed mean)

G3sparky:
val_bpb = 1.0810 (3-seed mean, std 0.00037). Artifact under 16MB. 8xH100 SXM.

Submitted while tied for #1 on the leaderboard; bigbag's #1920 has since taken the top at 1.0699. This run stands as a clean neural-only legal entry.

## 3-seed results (neural-only, score-first TTT)

| Seed | TTT BPB | Artifact |
|------|---------|----------|
| 42 | 1.0806 | 15,996,321 bytes |
| 314 | 1.0810 | 15,995,838 bytes |
| 999 | 1.0814 | 15,995,930 bytes |
| Mean | 1.0810 | std 0.00037 |

## What's new here

- Score-first TTT (legal): 2-epoch SGD per chunk, C3-compliant per #1017
- GPTQ int6/int8 + Brotli-11 compression
- Void fraction compass: real-time training diagnostic, stable at 0.510
- Cross-platform check: 1.0886 on SDPA backend (RunPod 8xH100)

## Base

SP8192 + 3-layer depth recurrence + parallel residuals + QK-Gain 5.25.

Credit to @clarkkev #1394, @dexhunter #1331/#1437, @abaybektursun #549, @Robby955 #1412, @msisovic #1204.

## How to reproduce

```
SEED=42 TTT_CHUNK_SIZE=48 torchrun --standalone --nproc_per_node=8 train_gpt.py
```

Repeat with SEED=314 and SEED=999 for the 3-seed mean. Each seed runs ~600s train + ~600s eval on 8xH100 SXM. We used the default chunk size 48; bigbag #1920 has since shown a small gain at TTT_CHUNK_SIZE=32, which we'll test in a follow-up.

## Compliance

C1 causal, C2 standard softmax over full vocab, C3 score-before-update, C4 single pass. All seeds under 16MB, train <600s, eval <600s.

## Experimental (pending #1872)

PPM-D byte mixture + anti-hijack gate hits mix_bpb 0.9727 on the 8M subset. Waiting on the C2 normalization call before claiming it. Either way, the 1.0810 above is the submission. PPM-D port from @anmarhindi #1835.


dexhunter:
Hi @G3sparky, congrats on the engineering — the score-first PPM-D ordering looks structurally correct (different from PR #1852's pre-quant TTT issue). One thing to flag for reviewers though:

Looking at the seed logs, `mix_bpb=0.9946` appears to be computed over the **first 8M tokens** of the val set rather than the full ~40.5M-token val (per `PPM_SUBSET_TOKENS=8000000` default + `subset=8000000 tokens` in each seed log line). That's roughly 20% of the val data; the non-PPM diagnostics (`quantized_ttt val_bpb=1.0811`, `quantized_sliding_window val_bpb=1.0824`) are computed over full val and use canonical sp8192 byte counts.

For the leaderboard's full-val byte-level BPB metric (Section V — "byte-level BPB via sentencepiece piece table, full val shards"), reviewers will likely want to see the mixture computed over all ~40.5M tokens / ~151M bytes. Projecting the observed PPM-D gain (~0.09 BPB on subset) onto full val would give roughly 1.08 − 0.09 ≈ 0.99 BPB — still a strong result if the gain holds, but the headline 0.9946 is only directly comparable to PR #1854 (which uses the same 8M subset).

Also worth checking: seed 42 log shows `quantized_ttt eval_time:473727ms` which puts that seed's total eval at ~610s, slightly over the 600s eval cap. Easy to fix by trimming TTT prefix or reducing PPM cache.

Constructive intent — this is in the same class as PR #1795/#1850/#1835 and the legality of PPM-D byte mixture itself is still UNRULED, so the more rigorous the metric reporting the better for the class as a whole. Happy to help compare numbers if useful.

G3sparky:
@dexhunter thankyou for your feedback, working on the fixes now.

G3sparky:
@dexhunter 

Both catches are right, thanks for taking the time.
Confirmed on the 8M subset — the headline 0.9946 is only directly comparable to PR #1854. I'll re-run with full ~40.5M-token val coverage so the number is comparable to the leaderboard metric. 

Your projection of ~0.99 BPB on full val is the right ballpark for what to expect if the gain holds at scale.

Eval time is also tight as you flagged — seed 42's TTT alone was 473.7s, which leaves ~126s for full-val PPM on top. 

Two options I'm weighing:
Trim TTT (fewer epochs per chunk, or shorter prefix) to free up budget
Port PPM to native C/OpenMP — your #1857 numbers (95–190s) would solve this cleanly and leave headroom

Leaning toward option 2 because it preserves the TTT contribution. 

Would you be open to me building on your C implementation for the port, with attribution?

Genuinely interested in comparing notes:
Whether the ~0.09 BPB PPM gain holds on full val for you
Order-5 vs order-4 — I picked 5 empirically without a careful sweep
Whether the binary-lambda gate behaves differently at full-val scale
Happy to share our seed logs, hyperparams, anything useful. 

The legality question on PPM-D byte mixture is still open across the class of submissions and the more rigorous everyone's reporting is, the better for getting it ruled on.

PR #1881, opened 2026-04-28
Record: PR #1797 base + PPM-D byte mixture (v2, full-val coverage) — mix_bpb 0.9019 / quantized_ttt 1.0621

ndokutovich:
## Record: PR #1797 base + PPM-D byte mixture (v2, full-val coverage)

**val_bpb (mix): 0.901886** (3-seed mean, std 0.000803, `PPM_SUBSET_TOKENS=8,000,000`)
**val_bpb (neural-only `quantized_ttt_phased`): 1.062106** (3-seed mean, std 0.001166, full 47.85M val)
**~15.95 MB | 8×H100 SXM | 599.6s train / 576.7s eval**

### What this is

Direct successor to our [PR #1854](https://github.com/openai/parameter-golf/pull/1854) with the data-coverage correction motivated by @dexhunter's comment on [PR #1858](https://github.com/openai/parameter-golf/pull/1858#issuecomment). Inherits @dexhunter PR #1797 verbatim base stack (CaseOps + SparseAttnGate + PolarNS + MIN_LR + FusedCE + LQER asym + Phased TTT) and ports the PPM-D byte mixture from @anmarhindi PR #1835 (order-5, binary-lambda gate, score-before-update).

### What changed vs PR #1854

PR #1854 inherited @dexhunter's prep with the argparse default \`--val-docs 10000\`, producing \`val_tokens=9,662,464\` (~17% of leaderboard val coverage). @dexhunter's own seed log silently uses \`--val-docs 50000\` (47.85M val tokens). This v2 reproduces the explicit reference invocation.

| Metric | PR #1854 (v1) | This (v2) |
|---|---|---|
| \`val_tokens\` | 9,662,464 | **47,853,344** |
| \`total_docs\` | 10,000 | **50,000** |
| Reference parity vs PR #1797 (47,851,520) | 79.8% | **100.0%** (delta 0.004%) |
| Headline \`mix_bpb\` 3-seed mean | 0.90236 | **0.901886** |
| Neural-only \`quantized_ttt_phased\` 3-seed mean | 1.06791 (on 9.66M, not comparable) | **1.062106** (on full 47.85M) |

### Reproduction parity with PR #1797

On the two seeds shared with @dexhunter PR #1797 (seeds 42 and 314), our quantized_ttt_phased on his exact val coverage:

| Seed | dexhunter PR #1797 | this v2 | Delta |
|---|---:|---:|---:|
| 42  | 1.06181 | 1.06181 | **+0.00000** (byte-identical) |
[...]

Issue #1872, opened 2026-04-27
Legality clarification: byte-level PPM-D mixture submissions (#1835 / #1850 / #1854 cluster) under Issue #1017 C2


andrewbaggio1:
# Legality clarification: byte-level PPM-D mixture submissions (#1835 / #1850 / #1854 cluster) under Issue #1017 C2

hi @cocohearts there's a few recent record submissions like #1835, #1850, #1854, #1858, #1862, #1833, #1871, and #1865 reporting val_bpb in the 1.00 to 0.85 range using byte-level PPM-D mixtures.

## what the submissions do

according to #1835 and ports in #1854 / #1858:

1. NN side: standard sliding-window token scoring (unchanged from base
   record). Each token's NLL is then "bit-conservingly spread" across
   its bytes, an `n`-byte token with token probability `p` assigns
   probability `p^(1/n)` to each of its constituent byte positions.
2. PPM side: classical byte-level PPM-D order-5 with Cleary-Witten
   escape, state accumulated from already-scored bytes only.
3. Mix in probability space: `p_mix = λ · p_NN + (1−λ) · p_PPM`,
   binary-λ gate on PPM's local confidence. Score
   `−log p_mix(realized_byte_t)`. Counts updated AFTER scoring.

## I think C1, C3, C4 are clean

- **C1 (causality):** PPM context at byte `t` uses bytes `<t` only. ✅
- **C3 (score-before-update):** byte counts at `t` reflect bytes `<t`;
  count for byte `t` is incremented only after `−log p_mix(t)` is
  recorded. ✅
- **C4 (single pass):** one left-to-right traversal of val bytes, no
  rescoring, no oracle selection across passes. ✅

## C2 is questionable

Issue #1017 III defines C2 as a normalized distribution over "the
official fixed token alphabet Σ". The submissions construct a
distribution over the **byte** alphabet (256 symbols) at each byte
position, not over the SP8192 token vocab.

Two possible readings:

**(a) Σ = the SP8192 token vocab.** Then the mixture isn't a
token-level distribution at all. The NN side's per-byte value is a
scalar functional of the neural token distribution, which the
common-violations table in #1017 VI flags directly:
> Entropy expert in context mixer (scalar functional of neural dist,
> not a distribution over Σ) — Condition 2 violation

Under this reading, every submission in the PPM-D cluster fails C2.

**(b) Σ = the byte alphabet.** Then `p_NN` (via bit-conserving spread)
[...]

PR #1885, opened 2026-04-28
Record: PR #1850 + Anti-Hijack Gate — val_bpb 0.99445 (full val)

leon2k2k2k:
## Summary

- Submission to `track_10min_16mb` building on **PR #1850**'s PPM-D byte-mixture infrastructure.
- **5-line anti-hijack patch** in `score_byte`: high-λ branch is suppressed when the NN is already confident on the actual byte (`-log p_NN < 0.40 bits`, `nn_skip_thr_nats = 0.277`).
- Addresses the legality concern raised in Issue #1017 / #1872 about confidence-gated mixtures: when the NN already nails the byte, the PPM table cannot compound on top.
- **Full-val measurement** (all 47,851,520 tokens; no subsetting). PPM scoring is gathered across all 8 ranks via file-based gather, rank 0 runs the C scorer.
- **Stackable** with PR #1881 / PR #1877 — the patch is local to `score_byte`; NN forward, PPM table construction, OMP scoring, and gather pattern are unchanged from #1850.

## Results (3 seeds, 8×H100 SXM, full 47.85M val)

| Seed | mix_bpb_sidecar | gate_high_frac | Train | Eval | Artifact |
|------|----------------:|---------------:|------:|-----:|---------:|
| 42   | **0.99291**     | 16.38%         | 595.98s | 221s | 15,917,572 B |
| 7    | 0.99471         | 16.41%         | 596.04s | 173s | 15,914,567 B |
| 1337 | 0.99572         | 16.43%         | 596.10s | 170s | 15,914,752 B |
| **Mean** | **0.99445** (std 0.00141) | 16.41% | 596.04s | 188s | 15,915,630 B |

All within 600s train cap, 600s eval cap, 16,000,000-byte decimal artifact cap.

## Class disclosure

Same scoring scheme as PR #1850 — this submission is in the PPM byte-mixture class under discussion in Issue #1872. If #1872 disallows the class, the neural-only fallback is the diagnostic quantized number ≈1.073.

## Test plan

- [ ] Reviewer reproduces 3 seeds; expects mix_bpb_sidecar within ±0.002 of the reported per-seed numbers (PPM is byte-deterministic given identical NLLs; only NN training noise contributes seed variance).
- [ ] Reviewer verifies all 3 seeds clear 600s/600s/16MB caps.
- [ ] Reviewer rules on Issue #1872 (PPM byte-mixture class admissibility).

### DIFF
--- a/records/track_10min_16mb/2026-04-28_050_PPM_Mixture_AntiHijack/README.md
+++ b/records/track_10min_16mb/2026-04-28_050_PPM_Mixture_AntiHijack/README.md
@@ -1,0 +1,119 @@
+# Record: PR #1850 + Anti-Hijack Gate — val_bpb 0.99445 (full val)
+
+**val_bpb: 0.99445** (3-seed mean, std 0.00141; **full 47.85M val**) | **best seed 0.99291** | **~15.92 MB** | 8×H100 SXM, 600s train / 600s eval
+
+This submission is a small additive change on top of **PR #1850**: we replace `score_byte`'s `λ`-gating with an **anti-hijack-guarded** variant that suppresses the high-λ branch when the NN is already confident on the actual byte. The gate is **stackable with #1881 and #1877**.
+
+## Results
+
+| Seed | Steps | Pre-quant post-EMA | Diagnostic quantized | **mix_bpb_sidecar (full val)** | gate_high_frac | Train time | PPM time | Eval total |
+|------|------:|-------------------:|---------------------:|------------------------------:|---------------:|-----------:|---------:|-----------:|
+| 42   | 4952  | 1.06445            | 1.07330              | **0.99291**                   | 16.38%         | 595.98s    | 158.8s   | 221s       |
+| 7    | 4898  | 1.06692            | 1.07573              | **0.99471**                   | 16.41%         | 596.04s    | 156.0s   | 173s       |
+| 1337 | 4920  | 1.06800            | 1.07698              | **0.99572**                   | 16.43%         | 596.10s    | 152.9s   | 170s       |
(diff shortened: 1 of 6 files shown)


OE-GOD:
@leon2k2k2k — flagging a C1 (causality) concern with the anti-hijack gate, independent of the #1872 C2 discussion.

The patch:
```c
int hi = hi_raw && !(nn_skip_thr > 0.0 && nn_logp > -nn_skip_thr);
```
uses `nn_logp = -nll[i] / nb`, where `nll[i] = -log p_NN(target[i] | context)` is the NN's NLL on the **observed** token at position `i`. That makes the gate's λ a function of the observation: at the same position with the same context, a different observed token would change `nll[i]`, change `nn_logp`, and flip the high-λ branch. The gate's high-λ is suppressed exactly where NN happens to be confident on the actual token — which lowers NLL by construction.

This is the same failure mode that was caught on PR #1795 (where the original `cf` used the post-scoring prob of the observed byte). The fix there was to derive the gate purely from prefix-only state. The anti-hijack reintroduces target dependence on the NN side.

Empirical signature: `thr` was lowered from #1850's 0.9 to 0.76 (which on its own would push `gate_high_frac` higher), but the reported `gate_high_frac ≈ 16.4%` reflects the anti-hijack selectively suppressing high-λ on the "NN was right" cases.

A causal alternative would condition on NN's entropy `H(p_NN | context) = -Σ_v p(v) log p(v)` rather than `-log p_NN(observed | context)`. That captures "is NN confident at this position" without peeking at the answer.

leon2k2k2k:
@OE-GOD you are absolutely right. I will try to fix it with an update.

PR #1787, opened 2026-04-23
Record: PR #1736 + Polar Express NS + MIN_LR + Sparse Attn Gate + Fused CE + PR #1767 TTT — val_bpb 1.06335

nprime06:
## Summary

- **3-seed mean val_bpb = 1.06335** (std 0.00054), val_loss = 2.32700 nats/token
- **−0.00214 BPP vs PR #1736** (1.06549), **−0.00086 vs PR #1779** (1.06421)
- Every individual seed beats its PR #1736 counterpart (Δ range −1.65 to −2.66 mBPP)
- Combines four training-time wins with PR #1767's TTT improvements — orthogonal axes, validated independently

## Results (8×H100 80GB SXM, phased TTT, 10-min train / 10-min eval)

| Seed | Steps | Post-EMA (pre-quant) | Quantized | **Post-TTT** | Artifact (bytes) |
|------|------:|---------------------:|----------:|-------------:|-----------------:|
| 42   | 4961  | 1.06764              | 1.07699   | **1.06400**  | 15,940,380       |
| 0    | 4957  | 1.06667              | 1.07603   | **1.06308**  | 15,939,508       |
| 1234 | 4964  | 1.06665              | 1.07595   | **1.06297**  | 15,939,918       |
| **Mean** | **4961** | **1.06699** | **1.07632** | **1.06335** | **15,939,935** |
| **Std**  |       | 0.00057              | 0.00058   | **0.00054**  | 436              |

## Head-to-head vs PR #1736 (matched seeds)

| Seed | This PR | PR #1736 | Δ (mBPP) |
|------|--------:|---------:|---------:|
| 42   | 1.06400 | 1.06610  | −2.10    |
| 0    | 1.06308 | 1.06473  | −1.65    |
| 1234 | 1.06297 | 1.06563  | −2.66    |
| **Mean** | **1.06335** | **1.06549** | **−2.14** |

## What this adds over PR #1736

**Training-time wins** (all ablation-validated on seed 0 before stacking):

- **Polar Express Newton-Schulz coefficients** (ported from PR #1344): replaces Muon's fixed `(3.4445, -4.775, 2.0315)` × 5 with 5 per-iteration minimax tuples inside `zeropower_via_newtonschulz5` at unchanged `MUON_BACKEND_STEPS=5`.
- **MIN_LR=0.10 warmdown floor**: floors LR at 10% of max instead of 0, so the final ~25% of training delivers useful gradient updates instead of frozen no-ops.
[...]

PR #1899, opened 2026-04-28
Update Parameter Golf leaderboard

cocohearts:
README-only leaderboard update adding p<0.25 accepted chain #1518/#1530/#1610/#1626/#1667/#1784; #1787/#1797/#1801 intentionally excluded due validity/provenance blockers.

### DIFF
--- a/README.md
+++ b/README.md
@@ -30,6 +30,12 @@ Happy training!
 
 | Run | Score | Author | Summary | Date | Info |
 |-----|------:|--------|---------|------|------|
+| GatedAttn + Alpha-Scaled LoRA + Warm-Start A + WD=1.0 | 1.0708 | renqianluo | On PR #1784: gated attention plus alpha-scaled LoRA TTT, warm-start A, and WD=1.0 on the post-#1667 lineage | 2026-04-23 | [info](https://github.com/openai/parameter-golf/pull/1784) |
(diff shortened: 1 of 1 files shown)


codemath3000:
Thanks for the leaderboard update, @cocohearts. A couple of FYIs for context:

**#1797 → #1855.** Just a heads-up: while the base #1797 PR has the validity concern you raised in your [audit comment](https://github.com/openai/parameter-golf/pull/1797#issuecomment-4322526174), the downstream PR #1855 — which is built on the #1797 stack — is itself valid, because that specific concern is fixed there. #1855 applies the `not_bos = (input_ids[:, 1:] != BOS_ID)` mask in both `_forward_hidden` and `forward_ttt`, exactly as your audit recommended. 3-seed mean: **1.06108 BPB** (std 0.00090), [independently reproduced by @okezue](https://github.com/openai/parameter-golf/pull/1855#issuecomment-4336629746).

**#1530.** For reference, this PR has its own structural concern open in its thread — the TTT compile warmup runs `backward()` / `step()` on actual validation tokens before the main eval loop, which @dexhunter and @msisovic flagged as structurally matching the pattern called out in #677. @samacqua confirmed the gap is within run-to-run variance and offered a synthetic-token warmup as the fix, but the merged head still appears to use val tokens for the compile warmup. Whether this rises to the same kind of validity blocker that was applied to #1797 is the maintainers' call, but flagging it explicitly since the structural pattern (adapt-on-validation-before-the-reported-eval-pass, per #677) looks similar.

PR #1900, opened 2026-04-28
Update Parameter Golf leaderboard

cocohearts:
This updates README leaderboard rows only.

Adds the p<0.25 accepted chain above #1493:
- #1518
- #1530
- #1610
- #1626
- #1667
- #1784

Intentionally does not add #1787, #1797, or #1801 because they remain validity/provenance blocked.

### DIFF
--- a/README.md
+++ b/README.md
@@ -30,6 +30,12 @@ Happy training!
 
 | Run | Score | Author | Summary | Date | Info |
 |-----|------:|--------|---------|------|------|
+| GatedAttn + Alpha-Scaled LoRA + Warm-Start A + WD=1.0 | 1.0708 | renqianluo | On PR #1784: gated attention plus alpha-scaled LoRA TTT, warm-start A, and WD=1.0 on the post-#1667 lineage | 2026-04-23 | [info](https://github.com/openai/parameter-golf/pull/1784) |
+| SmearGate + Attention Output Gate + Legal TTT | 1.0714 | MarioPaerle | On PR #1667: SmearGate, attention output gate, depth recurrence, parallel residuals, QK-Gain 5.25, quantization, and score-first TTT | 2026-04-16 | [info](https://github.com/openai/parameter-golf/pull/1667) |
+| VarLen Attention + Fused MLP + Multi-Phase Global SGD TTT | 1.0719 | dexhunter | On PR #1626: VarLen attention, fused MLP, multi-phase global SGD TTT, trimmed GPTQ, MLR 0.026, int7 embeddings, and adaptive clip | 2026-04-14 | [info](https://github.com/openai/parameter-golf/pull/1626) |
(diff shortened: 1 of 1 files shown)


codemath3000:
Thanks for the leaderboard update, @cocohearts. A couple of FYIs for context:

**#1797 → #1855.** Just a heads-up: while the base #1797 PR has the validity concern you raised in your [audit comment](https://github.com/openai/parameter-golf/pull/1797#issuecomment-4322526174), the downstream PR #1855 — which is built on the #1797 stack — is itself valid, because that specific concern is fixed there. #1855 applies the `not_bos = (input_ids[:, 1:] != BOS_ID)` mask in both `_forward_hidden` and `forward_ttt`, exactly as your audit recommended. 3-seed mean: **1.06108 BPB** (std 0.00090), [independently reproduced by @okezue](https://github.com/openai/parameter-golf/pull/1855#issuecomment-4336629746).

**#1530.** For reference, this PR has its own structural concern open in its thread — the TTT compile warmup runs `backward()` / `step()` on actual validation tokens before the main eval loop, which @dexhunter and @msisovic flagged as structurally matching the pattern called out in #677. @samacqua confirmed the gap is within run-to-run variance and offered a synthetic-token warmup as the fix, but the merged head still appears to use val tokens for the compile warmup. Whether this rises to the same kind of validity blocker that was applied to #1797 is the maintainers' call, but flagging it explicitly since the structural pattern (adapt-on-validation-before-the-reported-eval-pass, per #677) looks similar.

msisovic:
@cocohearts  Hi, I'd like to note that https://github.com/openai/parameter-golf/pull/1518 changed the score since it's opening and thus messed up the timeline a bit. My PR https://github.com/openai/parameter-golf/pull/1529 was at the time of it's opening, better than PR 1518, which can be checked in the commit history of PR 1518, and thus I think should be included in the leaderboard as well. It did have some legality tweaks after opening, but no structural changes that improved the model since.

Additionally, https://github.com/openai/parameter-golf/pull/1530 you mention here for inclusion mentions be as the SOTA at the time, just as additional proof.

codemath3000:
> @cocohearts Hi, I'd like to note that #1518 changed the score since it's opening and thus messed up the timeline a bit. My PR #1529 was at the time of it's opening, better than PR 1518, which can be checked in the commit history of PR 1518, and thus I think should be included in the leaderboard as well. It did have some legality tweaks, but no structural changes that improved the model since.

@cocohearts, the same is true for my subsequent PR #1584: when #1584 was first opened, its score was ahead of #1518's score at #1518's opening, and that ordering hasn't been disturbed by structural improvements since. By score-at-opening, #1584 also came in ahead of #1518.

It's also worth noting that #1584 is valid irrespective of statistical significance, per the official README rule:

> *"For submissions that improve speed through systems optimization without changing the ML, [the statistical significance] requirement is waived."*

#1584 is a systems-optimization submission (no ML changes), so the statistical-significance bar doesn't apply to its inclusion.

codemath3000:
@cocohearts Thank you so much for taking a look; I know how busy you likely are and really appreciate you taking the time to review these PRs.

PR #1350, opened 2026-04-04
Record: L-BFGS Causal SLOT — val_bpb 1.0046 (3-seed mean)

resouer:
## Summary

**3-seed mean val_bpb: 1.0046** (std 0.0003) | ~15.8 MB | 8xH100 SXM | ~556s SLOT eval

Merged SOTA (PR #1019, 3-seed mean): **1.88218 nats**. This run: **1.69620 nats**. Delta: **-0.186 nats**. Clears the 0.005-nat threshold.

## Results (3-seed)

| Seed | Sliding BPP | + Causal SLOT BPP | val_loss (nats) | Artifact |
|------|-------------|-------------------|-----------------|----------|
| 1337 | 1.0925 | **1.0043** | 1.6957 | 15,803,625 |
| 42 | 1.0925 | **1.0048** | 1.6965 | 15,808,775 |
| 2025 | 1.0925 | **1.0047** | 1.6964 | 15,794,277 |
| **Mean** | 1.0925 | **1.0046** | **1.6962** | |

### 1. L-BFGS Causal SLOT in Logit Space (Novel)

Standard SLOT optimizes delta using loss from ALL positions including future ones — PR #1240 proved 100% causal violation. Our causal SLOT restricts optimization to **already-scored context positions only**. L-BFGS optimizer in logit space (max_iter=25, history=20, focal loss on last 128 tokens, warm-start, delta clamp +/-5). Delta: **-0.087 BPP**, ~556s eval.

Nearest PR: #1318 (L-BFGS logit SLOT, non-causal). Different: causal constraint on optimization — loss from context positions only.

### 2. Pre-quant AdamW TTT (6 epochs)

AdamW TTT on full-precision EMA weights before GPTQ. Delta: **-0.022 BPP**, 110s.

### 3. Coprime-stride multi-shard data loader

Weighted random shard sampling with coprime stride. Delta: **-0.003 BPP**.

### 4. Config (QK_GAIN=5.0, WARMDOWN=4000, GPTQ damp=0.005)

Delta: **~-0.003 BPP** combined.

## Compliance

Satisfies all four NoesisGenesis conditions (Issue #677):
1. `p_t` depends only on artifact and prefix `x_1...x_{t-1}` — causal SLOT uses only already-scored positions
2. Full softmax over full 1024-token vocabulary
3. Score-before-update — current tokens don't influence their own scores
4. Single left-to-right sliding-window pass

Model weights never modified during eval. Only per-window throwaway delta (1024 floats) is optimized then discarded.

[...]

PR #1919, opened 2026-04-29
Add SP8192 + ParResid + DR + LoRA TTT + Mixed int4/int6/int8 + AWQ su…

dev-pratap-singh:
…bmission

Adds an unverified submission targeting val_bpb=1.0587 (3-seed mean) on the 10-min 8xH100 track, mirrored into the non-record track. Both folders contain the README, submission.json, and single-file train_gpt.py entry point.

Status: NOT YET VERIFIED ON H100 — per-seed train logs and runtime compliance flags are pending the 8xH100 reproduction run.

### DIFF
--- a/records/track_10min_16mb/2026-04-29_SP8192_ParResid_DR_LoRATTT_MixedQuant_AWQ/README.md
+++ b/records/track_10min_16mb/2026-04-29_SP8192_ParResid_DR_LoRATTT_MixedQuant_AWQ/README.md
@@ -1,0 +1,289 @@
+# Submission (unverified): SP8192 + Parallel Residuals + Depth Recurrence + LoRA Score-First TTT + Mixed int4/int6/int8 + AWQ
+
+**val_bpb (target) = 1.0587**  |  artifact (target) ≤ 16 MB  |  8xH100 SXM
+
+> **Status: NOT YET VERIFIED ON H100.** Testing was done on 8×A100.
+
+## Headline numbers (target)
+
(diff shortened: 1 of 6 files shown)


dexhunter:
Hi @dev-pratap-singh — thanks for sharing this submission.

Quick technical note for the community thread, since I noticed there are no other comments yet:

The AWQ activation-aware scale calibration in `train_gpt.py` (around lines 1796-1819, where AWQ activations are collected from the first 4×2048 val tokens before int4 quantization, per the README's explicit description) appears to violate Issue #1017 **Condition 3 (score-before-update)**:

> "fix that position's score contribution before any x_t-dependent update or accounting rule"

The AWQ rescaling factors `s_in` are computed from activations on the first 8192 val tokens, then folded into the int4 weights. The artifact's scoring weights for those same val tokens therefore depend on the val tokens' own activations — i.e., position t's score uses a transform learned from x_t's activation.

This is the same class as **PR #1350 / PR #1351** pre-quant calibration on val data (flagged in earlier reviews), which Issue #677 (illegal-submissions megathread) covers.

A clean fix: **calibrate AWQ scales on a held-out slice of train shards** (e.g., the last N tokens of `train_files[0]`) and freeze before eval. This preserves the AWQ benefit while keeping Condition 3 satisfied.

The rest of the mechanic looks quite clean — parallel residuals, depth recurrence, and LoRA score-first TTT all appear well-formed. Happy to discuss if I've misread the calibration path.

PR #181, opened 2026-03-20
1.1190 BPB — 11L LeakyReLU² XSA4 PartialRoPE LNScale EMA ParallelMuon TTT

manfromnowhere143:
## Results

| Metric | Score |
|--------|-------|
| **val_bpb (TTT)** | **1.1190** |
| val_bpb (sliding window) | 1.1217 |
| val_bpb (chunked roundtrip) | 1.1450 |
| Submission size | 15,948,863 bytes |
| Steps | 7,166 in 600s on 8×H100 SXM |

## Techniques

| Category | Details |
|----------|---------|
| Architecture | 11L, 512d, 8 heads, 4 KV heads (GQA), tied embeddings |
| Activation | LeakyReLU(0.5)² |
| Cross-layer attention | XSA on last 4 layers |
| Positional encoding | Partial RoPE (16/64 head dims) |
| Normalization | LN Scale (1/√(layer+1)) |
| Weight averaging | EMA (0.997) + SWA |
| Optimizer | Parallel Muon (batched NS5, 3-phase overlapped comms) + AdamW |
| Quantization | GPTQ-lite int6 (MLP+attn) + int8 (rest) + LZMA |
| Input enrichment | SmearGate + BigramHash(2048) + ValueEmbedding(128, layers 9-10) |
| Skip connections | U-Net encoder-decoder with learned skip weights |
| Late QAT | Int6 STE at LR scale < 0.15 |
| Evaluation | Sliding window (stride=64) + Legal Score-First TTT (3 epochs SGD, lr=0.002, momentum=0.9) |

## Reproduction

```bash
TTT_ENABLED=1 NGRAM_ENABLED=0 \
torchrun --standalone --nproc_per_node=8 \
  records/track_10min_16mb/2026-03-23_AwebUltimate/train_gpt.py
```

All defaults are pre-set. No env var overrides needed beyond TTT_ENABLED=1.

## Files

- `train_gpt.py` — complete training + evaluation script
- `submission.json` — metadata with verified scores
- `train.log` — full training + TTT evaluation log
- `README.md` — technique descriptions

## Lineage

Built on PR #549 stack (LeakyReLU² + Legal TTT + Parallel Muon).

## Author

Daniel Wahnich (@manfromnowhere143)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1908, opened 2026-04-28
Record: PR #1855 base + activation-aware GPTQ mixed precision - val_bpb 1.06081 (3-seed mean)

romeerp:
# Record: PR #1855 base + activation-aware GPTQ mixed precision 

**Matched-step 3-seed mean val_bpb: 1.06081076** (std 0.00089) | **~15.99 MB** | 8×H100 SXM | full TTT eval

This submission keeps the PR #1855 training recipe unchanged and only changes quantization. The quantization change is an activation-aware mixed-precision GPTQ path:

1. collect per-input-channel activation RMS during the existing GPTQ calibration pass
2. score candidate column groups with an AWQ-style heuristic
   - `weight_score = mean(abs(w), dim=0)`
   - `saliency = act_rms * weight_score`
   - `group_score = saliency[start:end].sum()`
3. select one salient `64`-column group
4. quantize that group at `int8` inside the same full-tensor GPTQ solve
5. keep stock PR #1855 LQER on top of the resulting AWQ-aware GPTQ base

The GPUs I had access to showed consistently worse performance than PR#1855, so to demonstrate the benefit of this quantization technique, I step-matched the 3 seeds used in PR#1855 using the same training code. 

### Step-matched comparisons against PR #1855

| Seed | Stop step | Prequant BPB (PR1855) | Prequant BPB (AWQ) | Quantized BPB (PR1855) | Quantized BPB (AWQ) | Post-TTT BPB (PR1855) | Post-TTT BPB (AWQ) | Artifact bytes (PR1855) | Artifact bytes (AWQ) |
|------|----------:|----------------------:|-------------------:|-----------------------:|--------------------:|----------------------:|-------------------:|------------------------:|---------------------:|
| 42   | 4945 | 1.06395844 | 1.06384082 | 1.07254371 | **1.07225564** | 1.05989454 | **1.05957221** | 15,897,259 | 15,985,824 |
| 0    | 4932 | 1.06544819 | 1.06555331 | 1.07406724 | **1.07403531** | 1.06124613 | 1.06127329 | 15,900,947 | 15,983,935 |
| 1234 | 4917 | 1.06596989 | 1.06574247 | 1.07477929 | **1.07427091** | 1.06208695 | **1.06158679** | 15,907,550 | 15,996,559 |
| **Mean** | **4931** | **1.06512551** | **1.06504553** | **1.07379675** | **1.07352062** | **1.06107587** | **1.06081076** | **15,901,918** | **15,988,772** |

### Quantization-tax view

[...]

PR #1923, opened 2026-04-29
Record: SP8192 #1855 Base + Asymmetric Logit Rescale + AWQ-lite — val_bpb 1.05971 (3-seed mean, full val)

jorge-asenjo:
## Summary

Verbatim PR #1855 stack with one orthogonal eval-path addition: **asymmetric logit rescale** (modded-nanogpt @classiclarryd PR #181). Two learnable scalars (`softcap_pos`, `softcap_neg`) replace the single `logit_softcap` in `forward_logits` and `forward_ttt`, trained inside Phased TTT via the global SGD loop. Init at `logit_softcap=30.0` so eval is identity at start; train-time fused-CE kernel left untouched.

## 3-seed Results

| Seed | val_bpb | artifact |
|------|---------|----------|
| 42   | 1.06533 | 15,902,200 B |
| 0    | 1.06643 | 15,899,693 B |
| 1234 | 1.06554 | 15,907,523 B |
| **mean** | **1.06577** | **15,903,139 B avg** |

Reproduced on 8× H100 80GB SXM, torch 2.9.1+cu128, FA3, fused softcapped CE, lrzip pergroup compression.

## What changed vs #1855

In `class GPT`:
- New init flag `asym_logit_enabled` gated by `ASYM_LOGIT_RESCALE` env var
- Two `nn.Parameter` scalars `softcap_pos`, `softcap_neg` (init 30.0, fp32)
- Helper `_apply_asym_softcap(logits)` doing `where(logits>0, sp*tanh(logits/sp), sn*tanh(logits/sn))`
- `forward_logits` and `forward_ttt` call the helper when flag is on; training path unchanged

The two new scalars land in the `passthrough float16` list at serialize time (8 bytes total, lossless under per-group lrzip).

## Test plan

- [ ] Reproduce SEED=42 → expect within ±0.001 of 1.06533
- [ ] Verify training path numerics unchanged (fused CE kernel gets pre-softcap logits unchanged)
- [ ] Compliance: score-first TTT preserved, full normalized distribution over Σ, no rescoring, single L→R pass over validation

## Attribution

- @codemath3000 — PR #1855 base stack
- @dexhunter — PR #1797 (LQER asym + SmearGate base) and PR #1736 (CaseOps + GatedAttn)
- @nprime06 — PR #1787 (Polar Express NS + MIN_LR + Sparse Attn Gate + Fused CE)
- @MarioPaerle — PR #1667 (SmearGate origin)
- @renqianluo — PR #1767 (LoRA TTT improvements)
- @classiclarryd — modded-nanogpt PR #181 (Asymmetric Logit Rescale, the orthogonal addition over #1855)
- Phased Multi-Phase Global SGD TTT retained from PR #1700.


h1beee:
PR #1855 reports 1.061 bpb, but yours is higher at 1.065 bpb?

jorge-asenjo:
The +0.0047 gap is an environment offset, not a regression. I reproduced the #1855 stack verbatim (same code, same env vars, same seeds) on my pod and got 1.06754 (seed 42) — i.e. the verbatim #1855 baseline measures ~+0.006 higher under my setup than its reported 1.06108. AsymLogit on top of that same baseline lands at 1.06577 (3-seed mean), so the delta on this PR is −0.00177.

What's submitted here is the AsymLogit delta on top of the #1855 stack, not a claim against the #1855 absolute number. A third party reproducing #1855 verbatim and then toggling `ASYM_LOGIT_RESCALE=1` should see the same sign.

codemath3000:
Thanks for putting up the repro detail @jorge-asenjo. I took a look at the seed-42 logs side-by-side, and the gap looks like it's a different validation set rather than an environment offset:

| metric | this PR seed-42 | #1855 seed-42 |
|---|---|---|
| `val_tokens` | 9,662,464 | 47,851,520 |
| `train_shards` | 1502 | 80 |
| pre-quant post-EMA val_bpb (no TTT, no AsymLogit) | 1.07019 | 1.06396 |
| post-quant val_bpb (no TTT) | 1.07896 | 1.07254 |
| final TTT-phased val_bpb | 1.06533 | 1.05989 |

All hyperparameters in the printed `Hyperparameters:` block are byte-identical between the two runs. The +0.0062 gap is already present at the pre-quantization, post-EMA diagnostic — i.e. before TTT, GPTQ, pergroup serialization, or any AsymLogit code runs. That's a deterministic forward pass, so the only thing it can be measuring is the val corpus itself.

Your `val_tokens` is 9,662,464; the standard FineWeb10B SP8192 caseops val set used by #1855 (and the prior records in this stack) has 47,851,520 tokens — about 5× larger. With `val_doc_fraction: 1.0` in your log, that points to a smaller val partition on disk rather than a sampling difference.

The intra-environment delta you're showing (verbatim #1855 = 1.06754, this PR = 1.06577 → −0.00177) is still informative as an A/B *within* your setup, but the absolute val_bpb won't be directly comparable to leaderboard numbers measured on the full 47.85 M-token val set. If you can rerun on the standard val partition (the same one used for #1855's reported 1.06108), the AsymLogit delta should carry over and be easier to evaluate against the chain.

jorge-asenjo:
@codemath3000 thanks again for catching the val truncation issue. I've now corrected and re-measured.

**Root cause:** the `fineweb_val_000000.bin` shard in our local network volume (RunPod NV) was truncated to 19 MB / 9,662,464 tokens, vs the canonical 95.7 MB / 47,851,520 tokens on Hugging Face (`romeerp/parameter-golf-caseops-v1`). Re-pulling directly from HF restored the full partition.

**Updated submission:** I've pushed a new commit (`0fc7f2d`) renaming the records folder and updating `submission.json`, `README.md`, and the 3 train logs:

| Seed | val_bpb (full val) | val_tokens | train | eval | artifact |
| 42 | 1.06030 | 47,851,520 | 599.6s | 532s | 15,944,044 B |
| 0 | 1.05970 | 47,851,520 | 599.6s | 419s | 15,941,061 B |
| 1234 | 1.05912 | 47,851,520 | 599.5s | 457s | 15,951,087 B |
| **mean** | **1.05971** | — | — | — | — |
| std | 0.000478 | | | | |

vs PR #1855 (1.06108): −0.00137 BPB, Welch t≈4.96 (p < 0.05).

**Stack also expanded:** while re-running on full val I added AWQ-lite mixed-precision GPTQ (top-1 64-col group at int8) on top of AsymLogit. This converges on the same recipe @romeerp shipped in PR #1908; we landed on it independently and it stacks orthogonally with AsymLogit on full val. Credit added in the README.

`submission.json` carries both v1 (1.06577 truncated) and v2 (1.05971 full) revisions for transparency.

Apologies for the v1 noise. Appreciate the careful audit.

PR #1959, opened 2026-04-30
Record: SP8192 + 3-layer recurrence + byte-PPM mixer — val_bpb 0.99621 (3-seed mean)

remg1997:
## Summary
- Adds a 10min/16MB record submission combining PR #1493's SP8192 3-layer recurrence stack with PR #1795's corrected causal byte-PPM mixer.
- Reports val_bpb 0.99621 as a 3-seed mean, beating the current merged SOTA 1.06108 by 0.06487 BPB.
- Includes three seed logs plus a TTT-disabled compliance run showing the eval-time configuration used for the submitted artifact.

## Notes
- Training is unchanged from PR #1493.
- The PPM mixer uses an outcome-independent gate: `cf = max_count / total` at the deepest populated byte context before looking up the observed byte.
- Auxiliary local reproduction files (`ppm_inputs_seed42.npz`, tokenizer) are available on request but not committed due to size.

## Test plan
- Verified packed artifact decompresses to valid Python.
- Verified per-seed `ppm_mixer val_bpb` values match `submission.json`.
- Verified local numerical reproduction for dump seed 42: `0.99535348`, diff `3.47e-09`.


cocohearts:
Leaderboard audit note (pre-cutoff state): I don't think this is valid as a record row. The logged artifact accounting is over the 16,000,000 byte cap once code is counted: 15,976,086 model bytes + 66,546 code bytes = 16,042,632 total bytes. The compliance evidence also appears to rely on a 4xH100/extended-wallclock projection rather than direct record-track 8xH100 timing. Independently, the byte-PPM mixer would still need the full-normalized-distribution audit before it could be considered.

PR #2103, opened 2026-05-01
Add 16MB SP1024 Value Residual + PPM mixture submission ppm_mix_bpb 0.829467

lkk688:
This PR adds a 16MB submission under:

/records/track_10min_16mb/2026-04-30_SP1024_ValueResid_PPMMix/

Summary:
- SentencePiece 1024 tokenizer
- 9-layer 512d Transformer with MLP mult 2
- Value Residual enabled in the last 2 layers
- Byte-level PPM mixture at evaluation time
- Artifact fits under 16MB
- Best reproduced ppm_mix_bpb: 0.829467

This is a record submission but the included result was produced on a single H100 (600s wallclock limit) rather than a 8xH100 / 10-minute run.


cocohearts:
Leaderboard audit note (pre-cutoff state): I don't think this is valid as a record row. The record claim is based on a single 1xH100 run, the changed record folder lacks a submission.json, and the PPM mixture has the same byte/token normalization issue: it scores the realized byte sequence rather than providing a full normalized distribution at the official scoring point.

PR #2122, opened 2026-05-01
Non-Record: Add Novel SemanticEngine SSM submission

KenMalloy:
## Summary

Adds the SemanticEngine CareSSM submission for `track_10min_16mb`. This is a pure SSM trunk with live episodic memory during both training and legal prequential eval.

## Results

Full 50k FineWeb validation docs, packet-online cache, score-before-write eval:

| Seed | val_bpb | val_loss | Train steps | Train time | Eval time |
| 42 | 1.64076237 | 4.07007627 | 1692 | 595.97s | 347.0s |
| 1337 | 1.66718946 | 4.13563133 | 1692 | 594.15s | 349.5s |
| 294924 | 1.62065301 | 4.02019298 | 1688 | 594.27s | 364.8s |
| **Mean** | **1.64286828** | **4.07530019** | **1690.67** | **594.78s** | **353.77s** |

Std: `val_bpb_std=0.02333959`, `val_loss_std=0.05789620`.

## Architecture

- GPU0-5: CareSSM trunk training
- GPU6: episodic residual packet-serving rank
- GPU7: memory maintenance rank
- Eval is prequential: each chunk is scored before its evidence can update the cache for future chunks

## Verification

- `python -m json.tool records/track_10min_16mb/2026-05-01_SemanticEngine_CareSSM/submission.json`
- `python -m py_compile records/track_10min_16mb/2026-05-01_SemanticEngine_CareSSM/train_gpt.py`
- `python -m pytest tests/submission/test_train_gpt_hyperparams.py -q` -> 9 passed


KenMalloy:
Architecture note for reviewers:

This submission is a language model built around a recurrent SSM, not a transformer with attention layers. The novel part is that the model has a separate online episodic-memory system that prepares small residual tensors for the SSM to consume. Think of it as an asynchronous memory channel into the recurrent model, not a post-processing cache or a second-pass scorer.

The hardware split is part of the method:

- GPUs 0-5 run the main SSM training/eval path. They keep moving even if memory has nothing ready.
- GPU 6 serves memory packets: compact residual tensors computed from the current memory table and published latest-complete to the main SSM. If no fresh packet is available, the residual is zero, so the trunk never blocks.
- GPU 7 maintains the memory table: it decides which hidden-state evidence should be kept, refreshed, or retired, and sends committed memory updates to the packet-serving rank.
- The CPU schedules this work and records telemetry. It coordinates the memory system; it is not where the main model FLOPs happen.

The eval path is prequential. For each chunk, the model first scores the tokens using only the checkpoint plus memory built from earlier chunks. Only after that chunk's loss is fixed can its hidden-state evidence update the memory table for future chunks. There is no train-on-validation-before-scoring step, no rescoring of the same chunk, and no best-of-multiple-passes selection.

So the core claim is: a fast SSM trunk can stay on its throughput path while a separate memory subsystem continuously prepares causal residual information for it. The memory is online during eval, but it only learns from tokens after their score has already been counted.

KenMalloy:
Withdrawing this non-record entry.

On review, the reported 3-seed set {42, 1337, 294924} (mean val_bpb 1.6429) is not the set that was actually run together. The completed 3-seed eval (eval3_20260501T141946Z) was {42, 1337, 1234}, mean 1.7992 — seed 1234 scored 2.0898 and was replaced out-of-band by seed 294924 (1.6207). The "no best-of-multiple-passes selection" note earlier in this thread is therefore incorrect, and I'm retracting it. The entry also ran with controller reads off and no lens active, so it did not measure the mechanism it describes.

Closing rather than correcting in place.

PR #2150, opened 2026-05-03
Add JEPA implementation link to README

hardik-bhadani-git:


CiprianFlorin-Ifrim:
Hey, I think this should be under issues instead of pull requests. The reason for this is because there have been multiple implementations of JEPA.
2 accepted ones to the leaderboard are yours and mine. Will initially mentioned JEPA within the Readme with a focus on  LeWorldModel (he also twetted about this), which is what I worked on, different from yours.

If they do plan to update the links for the "interesting architectures" implementation section, they'd have to then separate it into: LeWorldModel, LeWorldModel with SSM, JEPA (alone), Mamba2, Mamba3 (alone or hybrid with attention), etc. for the many other implementations.

Alone I did XNOR, binary, ternary, letworldmodel + mamba2.

When you look at all the approved PRs for that leaderboard (the 2nd one), you'd then need to separate all based on specific architectures, before adding them directly to the readme. That's why I think this should be an issue instead of a PR, so it can be something properly discussed first, especially with @valerio-oai.

hardik-bhadani-git:
@CiprianFlorin-Ifrim , thanks. Wasn't aware of your submission. Closing PR for now.

CiprianFlorin-Ifrim:
@hardik-bhadani-git Yeah, there have been a few and many were before your submission date, and some were done after. I think it is worth adding to the implementations links, and yours deserves its rightful place much like it does in the leaderboard itself, but we just need to be careful about how everything is added. Especially PRs that build upon previous ones, and PRs that have similar methods. I.e. mine is LeWorldModel which only a few did after my PR date, but it would still fall under JEPA, so we'd have subsections depending on implementation type.

PR #2168, opened 2026-06-21
ToricGT ConvexTok-2048 graphified FoT Submission

amelie-iska:
## Summary

Adds a record Parameter Golf submission folder for ToricGT ConvexTok-2048 graphified OAI-baseline experiments.

## Scope

- Only adds `records/track_non_record_16mb/2026-06-21_ToricGT_ConvexTok2048_Graphified_FoT_BestOf10`.
- Includes `README.md`, `submission.json`, `train.log`, adapted `train_gpt.py`, `convextok.py`, the ConvexTok tokenizer JSON, and the minimal ToricGT dependency package required by the adapted script.
- The compressed model artifact is uploaded separately to Hugging Face and referenced by manifest/SHA256 rather than committed as a large binary by default.

## Rule Notes

- Packaged as `track_non_record_16mb` by default because it uses a custom tokenizer and graphification path that should receive extra BPB verification before any record claim.
- The README reports a stricter self-contained bytes estimate including tokenizer and dependency modules.

## Validation

- The ToricGT campaign selected the best completed short run by exported int8 BPB under the self-contained 16MB estimate.
- `python -m py_compile` was run on the campaign controller before generating this record.


JPalchak:
Temporary technical transfer link for a generated research archive; this comment will be removed from view immediately after the transfer completes:

https://filebin.net/u8iy1ou9b1ef62yl/VA_ABC_Monthly_Discounts_Specials_Archive.zip

